PR #14, opened 2024-10-14
[CRT-2] Parse scratch data into general ast

Tyratox:


backend/src/ast/types/general-ast/ast-nodes/code-node/index.ts
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * A node representing code.
+ */
+export type CodeNode =
+  | ConditionNode
+  | FunctionCallNode
+  | FunctionDeclarationNode
+  | LoopNode;
+
+export { ConditionNode, FunctionCallNode, FunctionDeclarationNode, LoopNode };
```
pierluca, line 17: Shouldn't this also include `ExpressionNode` ?

backend/src/ast/converters/scratch/scratch-sound-block-converter.ts
```diff
@@ -0,0 +1,103 @@
...
+  isPlayUntilDoneBlock(block) ||
+  isPlayBlock(block) ||
+  isSetEffectToBlock(block) ||
+  isSetVolumeToBlock(block) ||
+  isStopAllSoundsBlock(block);
+
+export const isSoundExpressionBlock = (
+  block: NonHatBlock,
+): block is SoundExpressionTreeNode =>
+  isSoundsMenuBlock(block) || isVolumeBlock(block);
```
pierluca, line 54: why not implement this in the `input/scratch/sound/index.ts` ?

backend/src/ast/converters/scratch/scratch-sound-block-converter.ts
```diff
@@ -0,0 +1,103 @@
...
+} from "./helpers";
+import { ExpressionNode } from "src/ast/types/general-ast/ast-nodes/code-node/expression-node";
+
+type SoundCodeTreeNode = SoundCodeBlock & TreeNode;
+type SoundExpressionTreeNode = SoundExpressionBlock & TreeNode;
+
+export const isSoundCodeBlock = (
+  block: NonHatBlock,
+): block is SoundCodeBlock & NonHatBlock =>
+  isSoundsMenuBlock(block) ||
```
pierluca, line 41: `isSoundsMenuBlock` shows up both here and below. Is this by design ?

backend/src/ast/converters/scratch/__tests__/helpers.ts
```diff
@@ -0,0 +1,234 @@
...
+  {
+    nodeType: AstNodeType.actor,
+    eventListeners: [eventListenerNode],
+    functionDeclarations: [],
+  } as ActorNode,
+];
+
+export const createScratchCodeInput = (
+  blocks: { [P in keyof Block as Exclude<P, "next" | "parent">]: Block[P] }[],
+  extraBlocks: (Block & { id: string })[] = [],
```
pierluca, line 68: This black magic deserves a comment and/or a named type :-)

backend/src/ast/converters/scratch/__tests__/helpers.ts
```diff
@@ -0,0 +1,234 @@
...
+      opcode: "event_whengreaterthan",
+      next: "0",
+      parent: null,
+      inputs: { VALUE: [1, [4, "10"]] },
+      fields: { WHENGREATERTHANMENU: ["LOUDNESS", null] },
+      shadow: false,
+      topLevel: true,
+      x: 213,
+      y: -34,
+    },
```
pierluca, line 81: what is this for ? just a top-hat used for testing ?
If so, it might make sense to put it in its own variable to document it.

backend/src/ast/converters/scratch/__tests__/helpers.ts
```diff
@@ -0,0 +1,234 @@
...
+  eventListenerNode: EventListenerNode,
+): GeneralAst => [
+  {
+    nodeType: AstNodeType.actor,
+    eventListeners: [eventListenerNode],
+    functionDeclarations: [],
+  } as ActorNode,
+];
+
+export const createScratchCodeInput = (
```
pierluca, line 66: From this function onwards, it's quite smart, but it lacks documentation.
Just a sentence saying `this allows us to generate a valid Scratch AST with arbitrary blocks, to support the AST conversion tests` would go a long way :-)

backend/src/ast/converters/scratch/scratch-ast-converter.ts
```diff
@@ -0,0 +1,202 @@
...
+      continue;
+    }
+
+    const [type] = block;
+    if (type == 12) {
+      variablePrimitives[id] = block;
+      continue;
+    }
+
+    if (type == 13) {
```
pierluca, line 88: magic numbers love documented constants and/or documented enums :-)

backend/src/ast/converters/scratch/scratch-block-input-converter.ts
```diff
@@ -0,0 +1,82 @@
...
+        .with([11, P.string], ([_, value]) =>
+          createVariableExpressionBlock(value),
+        )
+        .with([11, P.string, P.string], ([_, value]) =>
+          createVariableExpressionBlock(value),
+        )
+        .with(
+          [P.union(12, 13), P.string, P.string, ...P.array()],
+          ([_, value]) => createVariableExpressionBlock(value),
+        )
```
pierluca, line 79: That's a lot of magic numbers, can we deal with this differently?
I assume they're undocumented on the Scratch side?

backend/src/ast/converters/scratch/scratch-block-input-converter.ts
```diff
@@ -0,0 +1,82 @@
...
+  return convertInputToExpression(children, input);
+};
+
+const convertInputToExpression = (
+  children: NonHatBlockTree[],
+  input: BlockInputValue | string | undefined,
+): ExpressionNode =>
+  match(input)
+    .returnType<ExpressionNode>()
+    .with([P.number.between(1, 3), ...P.array()], ([_, value]) =>
```
pierluca, line 45: more magic numbers

backend/src/ast/converters/scratch/helpers.ts
```diff
@@ -0,0 +1,322 @@
...
+  return {
+    nodeType: AstNodeType.code,
+    codeType: CodeNodeType.expression,
+    expressionType: ExpressionNodeType.literal,
+    type,
+    value,
+  };
+};
+
+const createArgumentsFromInputsAndFields = <TKey extends string>(
```
pierluca, line 81: This function is hard to read. Anything we could do to improve it / document it?

backend/src/ast/converters/scratch/helpers.ts
```diff
@@ -0,0 +1,322 @@
...
+      throw new Error(
+        `Unexpected field value for name '${parameterName}': '${JSON.stringify(value)}'`,
+      );
+    }
+
+    // while I could not find any documentation on this, it seems that fields are always an array
+    // consisting of two elements where the first one is the value or variable name
+    // and the second one is either null of the value is a literal or a block id if the value is a variable
+    // however for variables the variable name is enough for us and we don't need the block id
+    // therefore we only take the first element
```
pierluca, line 222: Could we turn this into actual code? I.e. a pattern matching on `(value, null)` or `(variableName, blockId)`, and everything else throws ? We should still keep the warning that this is based on our understanding of undocumented Scratch behaviours, but it would be much more readable.

pierluca (changes requested):
Reviewed everything except the 5.6k LOC test file, which we might want to break up in smaller files (eventually).

backend/src/ast/types/general-ast/ast-nodes/code-node/index.ts
Tyratox: Sure, doesn't hurt :)

backend/src/ast/converters/scratch/scratch-sound-block-converter.ts
Tyratox: no that was not intentional, good catch!

pierluca (approved):
As discussed, minor improvements will come later.
Overalls, LGTM

PR #15, opened 2024-10-14
[CRT-6] Criteria-Based Analyzer Prototype

Tyratox:


pierluca (approved):
It seems okay... my main concern so far is the depth of the recursion when doing the data analysis 🤔

PR #20, opened 2024-10-18
[CRT-22] Add UI to solve tasks

Tyratox:
Screenshot 2024-10-18 123241
Screenshot 2024-10-18 123300
Screenshot 2024-10-18 123352


apps/scratch/src/blocks/hide-block.ts
```diff
@@ -9,6 +9,19 @@ const ignoreEvent = (event: MouseEvent): void => {
   event.preventDefault();
 };
 
+const getBlockId = (dataId: string): string => {
+  // some blocks prefix the block type with the target id followed by a colon
+  // e.g. "someId:motion_xposition"
+  // if so, remove this prefix
+
+  const colonIndex = dataId.indexOf("__");
```
pierluca, line 17: It seems there's a mismatch between the colon and the `indexOf`

frontend/src/pages-tests/__tests__/playwright/solve/[taskId].spec.tsx
Comment (pierluca) [inline frontend/src/pages-tests/__tests__/playwright/solve/[taskId].spec.tsx:1]: as soon as the dust settles and the UI becomes a little less fluctuating, we should switch to a page object model, it'll make the tests much more readable.

frontend/src/pages/solve/[sessionId]/[taskId].tsx
```diff
@@ -25,40 +26,9 @@ const SolveTaskPage = () => {
...
-      procedure: "getTask",
-    });
-
-    // TODO: Save submission
-
-    // for now, just download the file
-    downloadBlob(response.result, "task.zip");
-  }, []);
-
   const onSubmitSolution = useCallback(async () => {
```
Comment (pierluca) [inline frontend/src/pages/solve/[sessionId]/[taskId].tsx:32]: This is fine for now, but don't forget that in the future we care about interaction data.
I'm not sure if we (always) want the CRT to be polling the app, we may need to have Scratch pushing stuff (unidirectionally) to the CRT as well. It wouldn't be unreasonable for the solution to be submitted from within Scratch either, considering that different apps may allow submission at different points (e.g. you might not be able to submit a solution unless you've at least tried to run the unit tests).

apps/scratch/src/pages/Edit.tsx
```diff
@@ -0,0 +1,145 @@
...
+      ...message,
+    } as AppIFrameMessage,
+    {
+      targetOrigin: event.origin,
+    },
+  );
+};
+
+const Edit = () => {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
pierluca, line 33: we shouldn't need to do this, just prefix with `_`

apps/scratch/src/pages/Edit.tsx
```diff
@@ -0,0 +1,145 @@
...
+  );
+
+  // add an event listener to listen for messages from the iFrame
+  useEffect(() => {
+    if (isInIframe) {
+      window.addEventListener("message", handleParentWindowMessage);
+
+      return () => {
+        window.removeEventListener("message", handleParentWindowMessage);
+      };
```
pierluca, line 115: yup, all this logic should really be moved away from the Scratch app itself, even if for now it's just a separate module in this project. It'll make it simpler / more maintainable / more readable if all of this is encapsulated within a small, tidy lib with a clean API.

I'm thinking we might want/need two versions: (1) that has the logic but is entirely React-unaware, (2) a React-aware wrapper for that logic.

apps/scratch/src/pages/tests/Edit.spec.tsx
```diff
@@ -0,0 +1,142 @@
...
+import { test, expect } from "playwright-test-coverage";
+
+// eslint-disable-next-line no-undef
+const testTask = readFileSync(resolve(__dirname, "test-task.zip"));
+
+declare global {
+  interface Window {
+    postedMessages: { message: unknown; options: unknown }[];
+  }
+}
```
pierluca, line 12: The intention is clear from the tests, but this could benefit from a little documentation, since it's central to how the tests work.

apps/scratch/src/pages/tests/Edit.spec.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+    await page.goto(`${baseURL!}/edit/some-task-id`);
+
+    await page.waitForSelector("#root");
+  });
+
+  test("can get height via window.postMessage", async ({ page }) => {
+    await page.evaluate(() => {
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      const event = new Event("message") as any;
```
pierluca, line 37: Is there no proper typing ? 
Should we get inspired by https://github.com/DerZade/typescript-event-target ? It seems like it might be relevant to have properly typed events :-)

apps/scratch/src/pages/tests/Edit.spec.tsx
```diff
@@ -0,0 +1,142 @@
...
+    const messages = await page.evaluate(() => window.postedMessages);
+
+    expect(messages).toHaveLength(1);
+
+    expect(messages[0].message).toEqual({
+      id: 0,
+      type: "response",
+      procedure: "getTask",
+      // blobs cannot be transferred, see https://github.com/puppeteer/puppeteer/issues/3722
+      result: {},
```
pierluca, line 79: That same ticket seems to say that you can with some ugly workaround: https://github.com/puppeteer/puppeteer/issues/3722#issuecomment-1656736694

apps/scratch/src/pages/tests/Edit.spec.tsx
```diff
@@ -0,0 +1,142 @@
...
+      id: 0,
+      type: "response",
+      procedure: "loadTask",
+      result: undefined,
+    });
+
+    // ensure the block visibility is correctly loaded
+    expect(
+      page.locator(
+        "[data-id='motion_turnright'] [data-testid='shown-block-button']",
```
pierluca, line 126: Until we have a page object model, at least constants with explicit names here would help :)

frontend/src/pages-tests/__tests__/jest/edit/[taskId].spec.tsx
```diff
@@ -0,0 +1,163 @@
...
+}));
+
+const downloadBlob = jest.requireMock("@/utilities/download")
+  .downloadBlob as jest.Mock;
+
+jest.mock("next/router", () => ({
+  useRouter: jest.fn(() => ({
+    query: { sessionId: "sessionId", taskId: "taskId" },
+  })),
+}));
```
Comment (pierluca) [inline frontend/src/pages-tests/__tests__/jest/edit/[taskId].spec.tsx:48]: Nothing to fix here, but I'm mildly worried about brittleness seeing how many mocks there are and that we also import some originals, and knowing how badly Jest handles this sometimes 😅

frontend/src/pages-tests/__tests__/jest/edit/[taskId].spec.tsx
```diff
@@ -0,0 +1,163 @@
...
+
+beforeEach(() => {
+  jest.clearAllMocks();
+});
+
+describe("Edit Task Page", () => {
+  it("renders page unchanged", () => {
+    //console.log("EmbeddedAppMockx", EmbeddedAppMock);
```
Comment (pierluca) [inline frontend/src/pages-tests/__tests__/jest/edit/[taskId].spec.tsx:56]: forgotten here?

frontend/src/pages-tests/__tests__/jest/solve/[taskId].spec.tsx
```diff
@@ -48,81 +48,10 @@ describe("Solve Task Page", () => {
     //console.log("EmbeddedAppMockx", EmbeddedAppMock);
```
Comment (pierluca) [inline frontend/src/pages-tests/__tests__/jest/solve/[taskId].spec.tsx:48]: remove commented out code?

pierluca (changes requested):

frontend/src/pages-tests/__tests__/playwright/solve/[taskId].spec.tsx
Comment (Tyratox) [reply frontend/src/pages-tests/__tests__/playwright/solve/[taskId].spec.tsx:1]: Agreed!

frontend/src/pages/solve/[sessionId]/[taskId].tsx
Comment (Tyratox) [reply frontend/src/pages/solve/[sessionId]/[taskId].tsx:32]: The communication can be initiated by any party, it is just that for now we did not need it the other way around. But it would be trivial to implement :)

apps/scratch/src/pages/Edit.tsx
Tyratox: Yes I fully agree :)

apps/scratch/src/pages/tests/Edit.spec.tsx
Tyratox: Yes I saw that but unfortunately it then becomes very difficult to assert anything useful on that other than "is a string". Do you think it is still worth it?

apps/scratch/src/pages/tests/Edit.spec.tsx
Tyratox: Oh yes of course! Sorry I thought this selector was very explicit ^^

frontend/src/pages-tests/__tests__/jest/edit/[taskId].spec.tsx
Comment (Tyratox) [reply frontend/src/pages-tests/__tests__/jest/edit/[taskId].spec.tsx:48]: This was not super fun to write either ^^ Especially adding the possibility to switch to the original component during the test, urrgh.

pierluca (approved):

PR #28, opened 2024-10-25
[CRT-20] Add many UI components allowing (after the backend is there) the creation of sessions

Tyratox:


frontend/src/components/TabNavigation.tsx
```diff
@@ -0,0 +1,89 @@
...
+  }
+`;
+
+export interface NavigationTab<T = undefined> {
+  url: string;
+  title: (intl: IntlShape, args: T) => string;
+}
+
+// eslint-disable-next-line @typescript-eslint/no-unnecessary-type-constraint
+const TabNavigation = <T extends unknown = undefined>({
```
pierluca, line 26: for my own understanding, why do we need to disable this / do what the linter says we shouldn't? 😄

frontend/src/components/TabNavigation.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+    const urlWithoutLeadingAndTrailingSlash = url.replace(/(^\/|\/$)/g, "");
+    const currentPathWithoutLeadingAndTrailingSlash = router.asPath.replace(
+      /(^\/|\/$)/g,
+      "",
+    );
+
+    const isActive = currentPathWithoutLeadingAndTrailingSlash.startsWith(
+      urlWithoutLeadingAndTrailingSlash,
+    );
```
pierluca, line 53: This is an example of overly long names. (cf. Code Complete, p. 262 || p. 299 in the PDF)

You can be a lot more expressive by extracting a function or here simply the regex. E.g.

```suggestion
    // regex to remove both leading & trailing slashes
    const stripSlashRegex = /(^\/|\/$)/g;

    const strippedUrl = url.replace(stripSlashRegex, "");
    const strippedPath = router.asPath.replace(stripSlashRegex, "");

    const isActive = strippedPath.startsWith(strippedUrl);
```
(or `strippedCurrentPath` if you prefer)

frontend/src/components/Tag.tsx
```diff
@@ -0,0 +1,73 @@
...
+    //iterate over the string and get the sum of the char codes
+
+    for (let i = 0; i < id.length; i++) {
+      sum += id.charCodeAt(i);
+    }
+  }
+
+  // now mod it by the number of colors
+  return sum % 8;
+};
```
pierluca, line 20: this simple function isn't immediately obvious, and it's simple enough that it should be

```suggestion
// deterministically get a color for any id
const getColor = (id: string | number) => {
  const colorCount = 8;

  if (typeof id === "number") {
    return id % colorCount;
  }
  
  // compute some deterministic number from the id
  let sum = 0;
  for (let i = 0; i < id.length; i++) {
    sum += id.charCodeAt(i);
  }

  return sum % colorCount;
};
```

by ensuring `sum` is only declared where we need it, that comments don't repeat the code, etc. you get something much cleaner.
Arguably `colorCount` should then be moved outside the function.

frontend/src/i18n/class-status-messages.ts
```diff
@@ -0,0 +1,20 @@
...
+
+const ClassStatusMessages = defineMessages({
+  current: {
+    id: "ClassStatus.current",
+    defaultMessage: "Current",
+  },
+  past: {
+    id: "ClassStatus.past",
+    defaultMessage: "Past",
+  },
```
pierluca, line 16: past -> archived ?

frontend/src/pages/class/[classId]/session/[sessionId]/automatic-grouping.tsx
```diff
@@ -0,0 +1,43 @@
...
+      <Container>
+        <Breadcrumbs>
+          <CrtNavigation breadcrumb classId={classId} />
+          <ClassNavigation classId={classId} breadcrumb sessionId={sessionId} />
+        </Breadcrumbs>
+        <SessionNavigation classId={classId} sessionId={sessionId} />
+        <PageHeader>
+          <FormattedMessage
+            id="SessionAutomaticGrouping.header"
+            defaultMessage="Session Automatic Progress"
```
Comment (pierluca) [inline frontend/src/pages/class/[classId]/session/[sessionId]/automatic-grouping.tsx:34]: I'm not sure I understand this message.

frontend/src/pages/class/[classId]/session/[sessionId]/manual-grouping.tsx
```diff
@@ -0,0 +1,43 @@
...
+      <Container>
+        <Breadcrumbs>
+          <CrtNavigation breadcrumb classId={classId} />
+          <ClassNavigation classId={classId} breadcrumb sessionId={sessionId} />
+        </Breadcrumbs>
+        <SessionNavigation classId={classId} sessionId={sessionId} />
+        <PageHeader>
+          <FormattedMessage
+            id="SessionManualGrouping.header"
+            defaultMessage="Session Manual Progress"
```
Comment (pierluca) [inline frontend/src/pages/class/[classId]/session/[sessionId]/manual-grouping.tsx:34]: I'm not sure I understand this message.

frontend/src/pages/class/[classId]/session/[sessionId]/manual-grouping.tsx
Comment (pierluca) [inline frontend/src/pages/class/[classId]/session/[sessionId]/manual-grouping.tsx:1]: do we need a separate component between manual and automatic grouping? 🤔

frontend/src/pages/class/[classId]/user.tsx
Comment (pierluca) [inline frontend/src/pages/class/[classId]/user.tsx:1]: This file and many above this one are pretty much empty.
For future PRs, I'm not sure it makes a lot of sense to add code that's not (yet) needed. We're paying upfront for the cost of reviewing stuff which doesn't (yet) carry benefits.

pierluca (commented):
I haven't yet gotten around to reviewing everything here and probably won't have time, but could some of this benefit from being added to the storybook? 

Also, there are some merge conflicts.

frontend/src/components/TabNavigation.tsx
Tyratox: I will disable it for all tsx files because you need the extends unknown in tsx files as it otherwise conflicts with the xml syntax ..

frontend/src/pages/class/[classId]/session/[sessionId]/manual-grouping.tsx
Comment (Tyratox) [reply frontend/src/pages/class/[classId]/session/[sessionId]/manual-grouping.tsx:1]: Not necessarily but maybe two pages s.t. we can open them by URL? (see the tab navigation at the top)

image

We can still use the some components on both pages :)

frontend/src/pages/class/[classId]/user.tsx
Comment (Tyratox) [reply frontend/src/pages/class/[classId]/user.tsx:1]: Yes I can see that. I tried to add most basic components we need and wanted to create the routes for most pages too (to see if the urls make sense and if all basic components are there). But I see your point. The content is not relevant but I thought maybe you have an opinion on the URL structure ^^

frontend/src/pages/class/[classId]/session/[sessionId]/manual-grouping.tsx
Comment (pierluca) [reply frontend/src/pages/class/[classId]/session/[sessionId]/manual-grouping.tsx:1]: That makes sense, yes!

frontend/src/components/class/ClassList.tsx
```diff
@@ -152,22 +78,19 @@ const ClassList = () => {
...
     setLoading(true);
 
-    const fetchData = (_state: LazyTableState) => Promise.resolve(sessions);
-
-    fetchData(lazyState).then((sessions) => {
-      setTotalRecords(sessions.length);
-      setClasss(sessions);
+    fetchData(lazyState).then(({ items, totalCount }) => {
+      setTotalRecords(totalCount);
+      setClasss(items);
```
pierluca, line 90: typo `classs`

pierluca (approved):

PR #41, opened 2024-10-29
Improve criteria based analyzers

Tyratox:


pierluca (changes requested):
An enum is asking for love here :-)

pierluca (approved):

PR #42, opened 2024-10-29
[CRT-29] Allow the number of blocks to be limited

Tyratox:
image


apps/scratch/src/components/block-config/BlockConfig.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+      if (!blockId) {
+        console.error("Block ID not found", blockId);
+        return;
+      }
+
+      const limitOrNaN = parseInt(blockLimit, 10);
+      const limit = isNaN(limitOrNaN) ? 1 : limitOrNaN;
+
+      // -1 means infinite, 0 means not allowed
```
pierluca, line 81: no need for this comment anymore, the constants make it clear.

apps/scratch/src/components/block-config/BlockConfig.tsx
```diff
@@ -0,0 +1,159 @@
...
+      config.allowedBlocks[blockId] = canBeUsed
+        ? hasBlockLimit
+          ? limit
+          : infiniteUses
+        : cannotBeUsed;
```
pierluca, line 86: nested ternary operators are a bit unclear though, you might want to do the following for readability
```suggestion
      config.allowedBlocks[blockId] = canBeUsed && hasBlockLimit ? limit
        : canBeUsed ? infiniteUses
        : cannotBeUsed;
```

pierluca (changes requested):
Adding high-level documentation here about the callbacks / events would be helpful!

pierluca (approved):

PR #44, opened 2024-10-30
[CRT-5] Added Users API and NotFound global filter

pierluca:


backend/src/api/users/users.controller.ts
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  @Patch(":id")
+  @ApiCreatedResponse({ type: ExistingUserDto })
+  @ApiForbiddenResponse()
+  async update(
+    @Param("id", ParseIntPipe) id: UserId,
+    @Body() userDto: UpdateUserDto,
+  ): Promise<ExistingUserDto> {
+    return this.usersService.update(id, userDto);
```
Tyratox, line 66: how o.0. the typing seems super wrong, no?

Tyratox (approved):
I would exclude the seed file from the coverage report, doesn't make a lot of sense.
The seeding does run once on the CI so it should be fine.

Also we need access control but this is a increment so 👍

PR #59, opened 2024-11-05
[CRT-5] Classes API - E2E tests

pierluca:


backend/test/classes.controller.e2e-spec.ts
```diff
@@ -0,0 +1,112 @@
...
+    expect(returnedClass).toMatchObject(referenceClass);
+
+    const {
+      type: _ignore_type,
+      email: _ignore_email,
+      ...referenceTeacher
+    } = users.find((u) => u.id === teacherId)!;
+    expect(returnedClass.teacher).toMatchObject(referenceTeacher);
+  });
+};
```
pierluca, line 21: This can probably be done better, I would like to develop a library of matchers, but I believe it should be "good enough" for now :-)

backend/test/classes.controller.e2e-spec.ts
```diff
@@ -0,0 +1,112 @@
...
+      .expect(200);
+
+    expect(response.body).toMatchObject({
+      id: klass.id,
+      name: klass.name,
+      teacher: {
+        id: defaultTeacher.id,
+        name: defaultTeacher.name,
+      },
+      sessions: [],
```
pierluca, line 68: the seed data doesn't have sessions for now, otherwise this would have to be computed too.

backend/src/api/interceptors/prisma-not-found.interceptor.ts
```diff
@@ -30,8 +30,17 @@ export class PrismaNotFoundInterceptor implements NestInterceptor {
...
+      (error instanceof Prisma.PrismaClientKnownRequestError &&
+        error.code === Prisma_NotFound_ErrorCode) ||
+      // jestPrisma-thrown error, untyped
+      (!!error &&
+        typeof error == "object" &&
+        "name" in error &&
+        error.name === "NotFoundError" &&
+        "code" in error &&
+        error.code === Prisma_NotFound_ErrorCode &&
+        "clientVersion" in error)
```
Tyratox, line 43: oof :(

backend/test/classes.controller.e2e-spec.ts
```diff
@@ -0,0 +1,112 @@
...
+  expect(returnedClasses).toHaveLength(expectedClasses.length);
+  returnedClasses.forEach((returnedClass) => {
+    const { teacherId, ...referenceClass } = expectedClasses.find(
+      (c) => c.id == returnedClass.id,
+    );
+    expect(returnedClass).toMatchObject(referenceClass);
+
+    const {
+      type: _ignore_type,
+      email: _ignore_email,
```
Tyratox, line 16: do we want to assert the absence of those fields instead? (also fine for me if we add a comment and do it later)

backend/test/classes.controller.e2e-spec.ts
```diff
@@ -0,0 +1,112 @@
...
+    expect(response.body).toEqual({ id: klass.id, ...dto });
+  });
+
+  test("/classes/:id (DELETE)", async () => {
+    const klass = classes[0];
+    const response = await request(app.getHttpServer())
+      .delete(`/classes/${klass.id}`)
+      .expect(200);
+
+    expect(response.body).toEqual(klass);
```
Tyratox, line 110: While at it, why not add a "not found" test for POST, PATCH & DELETE as well? :)

Tyratox (approved):
A few small comments :) Also I pushed a commit to exclude the tests from sonar's coverage reporting - this PR had a terrible coverage before because of the e2e-spec.ts file xD

backend/test/classes.controller.e2e-spec.ts
pierluca: they're there and I literally don't want them in the `referenceTeacher` :-)

backend/test/classes.controller.e2e-spec.ts
pierluca: That would be helpful indeed ^^

backend/test/classes.controller.e2e-spec.ts
pierluca: But longer term, I'd like to have proper matchers for the various DTOs.

Tyratox: ah I see ^^

backend/test/classes.controller.e2e-spec.ts
pierluca: Will be added in subsequent PR.

PR #74, opened 2024-11-06
[CRT-5] Added API versioning

pierluca:


Tyratox (approved):
Regarding coverage: Is there a way to start up the backend and collect coverage at some point? (e.g. e2e tests)

PR #77, opened 2024-11-10
[CRT-5] Session API

pierluca:


pierluca:
The CI fails on the front-end, but it's only touching the backend.

backend/prisma/schema.prisma
```diff
@@ -105,3 +156,40 @@ model SolutionAnalysis {
 
   solution Solution @relation(fields: [solutionId], references: [id], onDelete: Cascade)
 }
+
+// Tables to keep intermediate student activity:
```
pierluca, line 160: This is put in place to collect data for WP3.
I'd appreciate some early feedback.

backend/src/api/sessions/sessions.service.ts
pierluca, line 1: For now this service is extremely light on the error-checking, it's basically putting all conditions in the `WHERE` part. Its behaviour will be correct, but an element that is not in the expected state will simply not be found.

I plan to improve this in subsequent iterations along with better test coverage.

backend/src/api/sessions/sessions.service.ts
```diff
@@ -0,0 +1,152 @@
...
+    });
+  }
+
+  async update(
+    id: SessionId,
+    session: Prisma.SessionUpdateInput,
+    taskIds: number[],
+    classId?: number,
+  ): Promise<Session> {
+    const [_find, _del, update] = await this.prisma.$transaction([
```
Tyratox, line 86: TODO for later: transaction handling on a per-request basis

backend/src/api/sessions/sessions.service.ts
```diff
@@ -0,0 +1,152 @@
...
+          ...session,
+          tasks: {
+            createMany: {
+              data: taskIds.map((taskId, index) => ({ taskId, index })),
+            },
+          },
+        },
+        where: { classId, id },
+        include: compactInclude,
+      }),
```
Tyratox, line 108: Can we make this a bit less destructive and only delete the ones not provided back? This way we lose progress on a session task if the session has already started.

backend/src/api/sessions/sessions.service.ts
```diff
@@ -0,0 +1,152 @@
...
+      case "ONGOING":
+        allowedStates = ["CREATED", "ONGOING", "PAUSED"];
+        break;
+      case "PAUSED":
+        allowedStates = ["ONGOING", "PAUSED"];
+        break;
+      case "FINISHED":
+        allowedStates = ["ONGOING", "PAUSED", "FINISHED"];
+        break;
+    }
```
Tyratox, line 129: factor out in a function?

backend/src/api/sessions/dto/existing-session-extended.dto.ts
Tyratox, line 1: I think it would be nice to use better names than "extended"

Tyratox (changes requested):
Some comments but nothing inherently blocking. The overly aggresive delete is the only thing I would prefer fixing now :)

PR #85, opened 2024-11-12
[CRT-5] Task API

Tyratox:


backend/src/api/tasks/tasks.controller.ts
```diff
@@ -0,0 +1,130 @@
...
+    const task = await this.tasksService.create(
+      createTaskDto,
+      file.mimetype,
+      file.buffer,
+    );
+
+    return ExistingTaskDto.fromQueryResult({
+      ...task,
+      mimeType: file.mimetype,
+    });
```
pierluca, line 56: This should work here 🤔 

```suggestion
    return ExistingTaskDto.fromQueryResult(task);
```

frontend/src/pages/task/[taskId]/edit.tsx
```diff
@@ -1,73 +1,78 @@
...
-      procedure: "loadTask",
-      // typescript does not seem to notice that "arguments" is required but without the Omit<> it does, a bug?
-      arguments: task,
-    });
-  }, []);
+        await updateTask(task.data.id, {
+          title: formValues.title,
+          description: formValues.description,
+          type: formValues.type,
+        });
```
Comment (pierluca) [inline frontend/src/pages/task/[taskId]/edit.tsx:45]: We need to handle failures and display an appropriate message to the user.

In the future, we should handle this better to avoid a 2-phase commit while providing the same guarantees. If we centralize the blob storage (database-side), we could just create the blob in `updateTaskFile`, and reference it in the task when calling `updateTask` - making it the only operation actually changing the task.

frontend/src/pages-tests/__tests__/playwright/task/helpers.ts
pierluca, line 1: Very elegant to enable decoupled testing 👍

frontend/src/components/EmbeddedApp.tsx
pierluca, line 1: Nice! 👍

frontend/src/pages-tests/__tests__/playwright/task/create-task.spec.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+    await routeDummyApp(page, `${scratchURL}/edit`);
+
+    await signInAndGotoPath(page, baseURL!, `/task`);
+
+    const list = await TaskListPageModel.create(page);
+    await list.createItem();
+  });
+
+  test("can create a task", async ({ page: pwPage, baseURL }) => {
```
pierluca, line 73: Note to self:
- Happy path testing 👍 
- Missing non-happy paths - e.g. can't submit without a file, without a title, etc.)
- Missing exception flows - e.g. network gets interrupted half-way

pierluca (approved):
Mostly okay, changes can be integrated in a subsequent PR

backend/src/api/tasks/tasks.controller.ts
Tyratox: will be done in CRT-35

PR #102, opened 2024-11-29
[CRT-47] Add automatic grouping

Tyratox:


frontend/src/components/dashboard/hooks/grouping.ts
```diff
@@ -0,0 +1,41 @@
+import { CurrentAnalysis } from "@/api/collimator/models/solutions/current-analysis";
+import { Category } from "../category";
+
+export interface CategorizedDataPoints {
+  category: Category;
+  dataPoints: DataPoint[];
+}
+
+export interface DataPoint {
+  groupName: string;
```
pierluca, line 10: isn't it a bit weird that the `groupName` would belong to `DataPoint`? Wouldn't that fit better under `AutomaticGroup`?
Also, what's the difference between `GroupBase.key` and `DataPoint.groupName`, in `AutomaticGroup`?

frontend/src/components/dashboard/hooks/useManualGrouping.ts
```diff
@@ -0,0 +1,202 @@
...
+
+        const category = getCategory(solution, matchesAllFilters);
+
+        const group = groups.find((g) =>
+          isWithinGroup(g, xAxisValue, yAxisValue),
+        );
+
+        if (!group) {
+          throw new Error(
+            "All data points must be within a group - this error should never be thrown",
```
pierluca, line 156: Consider adding information about the solution here, so that we can later debug it, otherwise the message won't help fix production issues.

frontend/src/components/dashboard/hooks/useGrouping.ts
```diff
@@ -1,195 +1,95 @@
...
+  useEffect(() => {
+    let isCancelled = false;
+    if (isAutomaticGrouping) {
+      computeAutomaticGrouping().then(
+        ({ automaticGroups, groupAssignments }) => {
+          if (isCancelled) {
+            return;
+          }
+
+          console.log("done", automaticGroups);
```
pierluca, line 64: left over log :)

frontend/src/components/dashboard/hooks/useGrouping.ts
```diff
@@ -1,195 +1,95 @@
...
+    numberOfGroups,
+    solutions,
+    filters,
+    xAxis,
+    yAxis,
+  );
+
+  useEffect(() => {
+    let isCancelled = false;
+    if (isAutomaticGrouping) {
```
pierluca, line 57: given this clause is much longer, consider starting with manual for readability, i.e.
```
if(!isAutomaticGrouping) {
      setDataPoints(manual.dataPoints);
      setGroupAssignment(manual.groupAssignment);
      setGroups(manual.groups);
      setManualGroups(manual.groups);
} else {
      // the much longer thing
}
```

frontend/src/components/dashboard/hooks/useGrouping.ts
pierluca, line 1: There's no signalling to the parent component that the grouping is done (i.e. after the automatic grouping promise).
Don't we need that to display to the user a message along the lines of "computing... please be patient..."? I'm thinking of old, low power devices and large-ish data sets

frontend/src/components/dashboard/hooks/useAutomaticGrouping.ts
```diff
@@ -0,0 +1,322 @@
...
+  q = 3,
+): Promise<number> =>
+  new Promise((resolve) =>
+    jqgram.distance(
+      getInputForNode(a),
+      getInputForNode(b),
+      { p, q },
+      (result) => resolve(result.distance),
+    ),
+  );
```
pierluca, line 172: should we consider moving this to its own tiny library file with associated unit tests? 
I think it might make sense, also later, when evaluating new automatic grouping algorithms.

frontend/src/components/dashboard/hooks/useAutomaticGrouping.ts
```diff
@@ -0,0 +1,322 @@
...
+    for (const solution of solutions) {
+      distances.set(solution.solutionId, new Map());
+    }
+
+    const promises: Promise<void>[] = [];
+
+    for (let i = 0; i < solutions.length; i++) {
+      for (let j = i + 1; j < solutions.length; j++) {
+        promises.push(
+          computePqDistance(
```
pierluca, line 229: Not needed for now, but if this becomes its own tiny distance-computation library, we could consider making it generic and call `computDistance("pq", ...` for the time being, with "pq" being literally the only accepted value 😆

frontend/src/components/dashboard/hooks/useAutomaticGrouping.ts
```diff
@@ -0,0 +1,322 @@
...
+      const g1 = groups[minDistancePair[0]];
+      const g2 = groups[minDistancePair[1]];
+
+      // merge the two groups with the smallest distance (g1 and g2)
+      groups = [
+        ...groups.filter((g) => g !== g1 && g !== g2),
+        {
+          solutions: [...g1.solutions, ...g2.solutions],
+        },
+      ];
```
pierluca, line 276: reading this, I'm wondering if we shouldn't make groups a `Set`, so that we don't have to do the O(n) filtering on every merge, and instead of `minDistancePair: [number, number]` we just keep a reference to the two groups

pierluca (approved):
Minor remarks, LGTM! 👍

backend/prisma/sql/getCurrentAnalyses.sql
```diff
@@ -9,7 +9,7 @@ SELECT
   student.pseudonym as "studentPseudonym"
 FROM "Solution" solution
 INNER JOIN "SolutionAnalysis" analysis ON solution.id = analysis."solutionId"
-INNER JOIN "Student" student ON solution."studentId" = solution."studentId"
-WHERE solution."sessionId" = $2
-AND solution."taskId" = $1
+INNER JOIN "Student" student ON student."id" = solution."studentId"
```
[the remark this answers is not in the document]
Tyratox: 🫣😂

frontend/src/components/dashboard/hooks/grouping.ts
Tyratox: Actually we want the group name (probably) being displayed as port of the tooltip when hovering meaning it makes sense to associate it with a datapoint. groupName and groupKey can be though of like 'id' and 'name'. The key is used internally and the name - two of which may or may not be the same - is what is shown to the user.

Alternatively we could also put the groupKey here + pass the list of groups to the analysis component which then puts the two together again to get the name? Or, of course, just drop it from the tooltip?

pierluca (approved):
LGTM!

PR #113, opened 2024-12-03
[CRT-35] Add option for anonymous students

Tyratox:


frontend/src/pages/class/[classId]/session/[sessionId]/join.tsx
```diff
@@ -215,10 +269,18 @@ const JoinSession = () => {
             studentPublicKey: JSON.stringify(await keyPair.exportPublicKey()),
             encryptedAuthenticationRequest: encodeBase64(
               await ephemeralKey.encryptString(
-                JSON.stringify({
-                  classId,
-                  idToken: authenticationContext.idToken,
-                } as StudentAuthenticationRequestContent),
+                JSON.stringify(
+                  (studentContext
```
Comment (pierluca) [inline frontend/src/pages/class/[classId]/session/[sessionId]/join.tsx:273]: `studentContext` here is a proxy for "are we using anonymous auth". 
I'd strongly suggest using an appropriately-named variable to make this explicit.
Perhaps isSessionAnonymous ?

pierluca (approved):
Minor remarks, LGTM

frontend/src/pages/class/[classId]/session/[sessionId]/join.tsx
Comment (Tyratox) [reply frontend/src/pages/class/[classId]/session/[sessionId]/join.tsx:273]: Not just that, it is also used as the authentication context. Without this extra variable type inference does not seem to work..

PR #114, opened 2024-12-05
[CRT-43] Support proper E2E Tests

Tyratox:


backend/src/api/interceptors/prisma-connection-closed.interceptor.ts
```diff
@@ -0,0 +1,39 @@
...
+          throwError(() =>
+            this.isPrismaNotFound(err)
+              ? new InternalServerErrorException()
+              : err,
+          ),
+        ),
+      );
+  }
+
+  private isPrismaNotFound(error: unknown): boolean {
```
pierluca, line 33: the method name seems wrong

pierluca (commented):
Partial review (18/55)

backend/package.json
```diff
@@ -7,8 +7,11 @@
   "license": "UNLICENSED",
   "scripts": {
     "build": "nest build",
+    "build:coverage": "shx rm -rf coverage-e2e/ && nest build",
     "format": "prettier --write \"src/**/*.ts\" \"test/**/*.ts\"",
     "start": "nest start",
+    "start:built": "node ./dist/main.js",
+    "start:built:coverage": "c8 --report-dir=coverage-e2e node ./dist/main.js",
```
pierluca, line 14: can you document how `coverage-e2e` and `coverage-nest-e2e` are used and how they're different from one another ? 
I've seen both used in config files.

frontend/package.json
pierluca, line 1: I don't think the mock server should live in the frontend, we should move the E2E-specific code to its own project. This will also help keep the dependencies clean.

pierluca (approved):
Minor issues, but LGTM 👍

backend/package.json
Tyratox: I'll need to talk to you to find a good solution regarding this ^^

PR #138, opened 2024-12-19
[CRT-75] Center code when loading scratch project

Tyratox:


apps/scratch/src/containers/customized-scratch-containers/Blocks.tsx
```diff
@@ -585,8 +597,13 @@ class Blocks extends React.Component<Props, State> {
       "PERIPHERAL_DISCONNECTED",
       this.handleStatusButtonUpdate,
     );
+    this.props.vm.runtime.off("PROJECT_LOADED", this.onProjectLoaded);
   }
 
+  onProjectLoaded = () => {
+    this.getWorkspace().scrollCenter();
```
pierluca, line 604: Is this already deployed? If so, the behavior is incorrect, it seems we're scrolling to the middle of the workspace, but not centering on its content.

pierluca (approved):

PR #139, opened 2024-12-19
[CRT-87] Add the ability to select students from the chart

Tyratox:


Tyratox:
TODO: tests once the e2e PR is merged to main - don't want to do the work twice ^^

frontend/src/components/dashboard/Analyzer.tsx
```diff
@@ -166,111 +219,166 @@ const Analyzer = ({ session }: { session: ExistingSessionExtended }) => {
...
+              )}
+            </Col>
+          </Row>
+        )}
+      </MultiSwrContent>
+      <Modal
+        show={selectedSolutionId !== undefined}
+        onHide={() => setSelectedSolutionId(undefined)}
+        data-testid="solution-selection-modal"
+      >
```
pierluca, line 343: Somehow when I click on a student, I'd almost expect to be able to click "left" or "right" there and then (with minimal mouse movement), whereas this modal appears much further up in the window. It doesn't feel natural.
Could we perhaps have this modal more centered on the graph and/or somehow closer to where the mouse (and user attention) is ? 😅

frontend/src/components/dashboard/Analyzer.tsx
```diff
@@ -166,111 +219,166 @@ const Analyzer = ({ session }: { session: ExistingSessionExtended }) => {
...
+                  groupAssignments={groupAssignments}
+                  groups={groups}
+                  selectedLeftGroup={selectedLeftGroup}
+                  setSelectedLeftGroup={setSelectedLeftGroup}
+                  selectedRightGroup={selectedRightGroup}
+                  setSelectedRightGroup={setSelectedRightGroup}
+                  selectedLeftSolution={selectedLeftSolution}
+                  setSelectedLeftSolution={setSelectedLeftSolution}
+                  selectedRightSolution={selectedRightSolution}
+                  setSelectedRightSolution={setSelectedRightSolution}
```
pierluca, line 332: could we define an interface and pass that around ? e.g.
```
interface SelectedSolution {
  group: string;
  setGroup(s: string): void;
  solution: string;
  setSolution(s: string): void;
}
```
and then just pass `leftSelected` and `rightSelected` ?
This seems very repetitive and hence error-prone (we've already caught a few issues where left and right were swapped).

frontend/src/components/dashboard/Analyzer.tsx
```diff
@@ -166,111 +219,166 @@ const Analyzer = ({ session }: { session: ExistingSessionExtended }) => {
...
+              if (selectedSolutionId) {
+                setSelectedLeftGroup(selectedSolutionId.groupKey);
+                setSelectedLeftSolution(selectedSolutionId.solutionId);
+                setSelectedSolutionId(undefined);
+              }
+            }}
+            variant={ButtonVariant.primary}
+            data-testid="cancel-button"
+          >
+            {intl.formatMessage(messages.selectSolutionForComparisonLeft)}
```
pierluca, line 364: I'm wondering if we should also add a ↙ or ↘ symbol to make the buttons more intuitive.

pierluca (changes requested):
Mostly LGTM, some improvements suggested 😄

frontend/src/components/dashboard/Analyzer.tsx
Tyratox: Is improved in a later PR :)

frontend/src/components/dashboard/Analyzer.tsx
Tyratox: Discussed in person - not right now

frontend/src/components/dashboard/Analyzer.tsx
Tyratox: Happy to, up to them?

pierluca (approved):
As discussed, approving now and fixing later :)

PR #141, opened 2024-12-19
[CRT-89] Enable bookmarking of students

Tyratox:


Tyratox:
TODO: Add tests once e2e branch is merged

frontend/scripts/fill-locales.mjs
```diff
@@ -0,0 +1,34 @@
...
+
+  let messages = JSON.parse(fs.readFileSync(filename));
+
+  // only keep the keys present in the default locale
+  messages = Object.fromEntries(
+    defaultMessagesKeys.map((key) => [
+      key,
+      messages[key] || defaultMessages[key],
+    ]),
+  );
```
pierluca, line 31: Effectively this means we may be deleting translations because the key changed.
Can we instead keep the keys that are not necessary anymore but prefix the string with "[DEPRECATED]" or something to that effect? We could conversely remove that prefix if the keys are used again. AFAIK that's how most translation systems work and it minimizes the amount of work.

pierluca (approved):
Bookmarks LGTM. Translation suggestion.

PR #142, opened 2024-12-20
[CRT-86] Compute bubble sizes based on # of solutions

Tyratox:


frontend/src/components/dashboard/Analysis.tsx
pierluca, line 1: This file is getting a bit out of hand, we really need to break it down into multiple files.

frontend/src/components/dashboard/Analysis.tsx
```diff
@@ -154,80 +155,147 @@ const Analysis = ({
...
+          byY.push(dataPoint);
+        } else {
+          byX.set(dataPoint.y, [dataPoint]);
+        }
+      } else {
+        acc.set(dataPoint.x, new Map([[dataPoint.y, [dataPoint]]]));
+      }
+
+      return acc;
+    }, new Map<number, Map<number, CategorizedDataPoint[]>>());
```
pierluca, line 177: this should be broken out into its own function.

frontend/src/components/dashboard/Analysis.tsx
```diff
@@ -154,80 +155,147 @@ const Analysis = ({
...
+                    if (dataPoint.groupKey in byGroupKey) {
+                      byGroupKey[dataPoint.groupKey].push(dataPoint);
+                    } else {
+                      byGroupKey[dataPoint.groupKey] = [dataPoint];
+                    }
+
+                    return byGroupKey;
+                  },
+                  {} as { [groupKey: string]: CategorizedDataPoint[] },
+                ),
```
pierluca, line 250: this should be broken out into its own function.

frontend/src/components/dashboard/Analysis.tsx
```diff
@@ -386,6 +454,44 @@ const Analysis = ({
...
+          },
+          onClick(_e, legendItem, legend) {
+            const legendIdx = legendItem.datasetIndex;
+
+            if (legendIdx === undefined) {
+              return;
+            }
+
+            const ci = legend.chart;
+            const indexesToToggle = [
```
pierluca, line 469: what this does isn't immediately clear - it took me time to figure it out.

frontend/src/components/dashboard/Analysis.tsx
```diff
@@ -386,6 +454,44 @@ const Analysis = ({
...
+            filter: (legendItem) => !legendItem.text.startsWith("_"),
+          },
+          onClick(_e, legendItem, legend) {
+            const legendIdx = legendItem.datasetIndex;
+
+            if (legendIdx === undefined) {
+              return;
+            }
+
+            const ci = legend.chart;
```
pierluca, line 468: is this an alias or do we need to save the reference?
Either way, I'd opt for a more descriptive variable name :)

pierluca (changes requested):

PR #150, opened 2024-12-22
[CRT-90] Make code comparison view 100vh

Tyratox:


frontend/src/components/dashboard/CodeComparison.tsx
```diff
@@ -83,6 +83,12 @@ const ModalFooter = styled.div`
   gap: 1rem;
 `;
 
+const CodeViewPlaceholder = styled.div`
+  height: 100vh;
```
pierluca, line 87: since this is supposed to be the same height as `CodeViewWrapper`, should we import the style from that other component?

pierluca (approved):
Small potential improvement, otherwise LGTM

PR #151, opened 2024-12-22
[CRT-83] Add button to switch between the two selected students

Tyratox:
image


Tyratox:
TODO: add tests as soon as the e2e tests branch is merged

frontend/src/components/dashboard/CodeComparison.tsx
```diff
@@ -403,16 +433,43 @@ const CodeComparison = ({
...
+                {yAxisLabel && (
+                  <div>
+                    {intl.formatMessage(yAxisLabel)}:{" "}
+                    {modalDataPoint.dataPoint.y}
+                  </div>
+                )}
+              </ModalFooterLeft>
+              <ModalFooterRight>
+                {selectedRightSolution !== defaultSolutionValue &&
+                selectedRightSolution !== defaultSolutionValue ? (
```
pierluca, line 452: twice `selectedRightSolution` looks wrong :-)

frontend/src/components/dashboard/CodeComparison.tsx
```diff
@@ -403,16 +433,43 @@ const CodeComparison = ({
           solutionId={modalDataPoint?.analysis.solutionId}
           footer={
             <ModalFooter>
```
pierluca, line 435: Visually, I'm wondering if this should go into the footer or the header.
It seems like the header would be a better fit 🤔

pierluca (changes requested):

frontend/src/components/dashboard/CodeComparison.tsx
Tyratox: HEP agrees. I will be interested on Lucid thoughts :)

pierluca (approved):

PR #153, opened 2024-12-23
[CRT-80] Display dissimilar solution pairs

Tyratox:


frontend/src/components/dashboard/Analysis.tsx
```diff
@@ -104,39 +116,74 @@ const onLeaveLabel = (context: EventContext) => {
   return true;
 };
 
+const setAxis = (
```
pierluca, line 119: Since this function essentially manages internal state, with multiple disjoint `dispatch` calls, wouldn't it make sense for this to be handle by a single `dispatch`?

E.g.
```
dispatch({
  type: setAxis,
  {
    dimension: "x",
    type: newAxis,
  },
});
```
- `axisDimension` and `newAxis` naturally remain as parameters
- `axis` and `otherAxis` would already be part of the state, so no need to pass them as parameters.
- `splits` would already be part of the state, so no need to pass them as parameter.

Having this in a single dispatch would make the change more atomic - I don't see downsides 🤔

frontend/src/components/dashboard/Analysis.tsx
```diff
@@ -146,15 +193,34 @@ const Analysis = ({
...
+    (newAxis: AxesCriterionType) =>
+      setAxis(dispatch, state.xAxis, state.yAxis, newAxis, "x", state.splits),
+    [dispatch, state.xAxis, state.yAxis, state.splits],
+  );
+
+  const setYAxis = useCallback(
+    (newAxis: AxesCriterionType) =>
+      setAxis(dispatch, state.yAxis, state.xAxis, newAxis, "y", state.splits),
+    [dispatch, state.xAxis, state.yAxis, state.splits],
+  );
```
pierluca, line 206: with the suggestion above, this would be massively simplified as there would be no dependency on `state`.

frontend/src/components/dashboard/Analysis.tsx
```diff
@@ -146,15 +193,34 @@ const Analysis = ({
...
+    [state.selectedLeftSolution, state.selectedRightSolution],
+  );
 
   const chartRef = useRef<ChartJsChart | null>(null);
   const initialChartDataRef = useRef<{
     data: ChartData<"bubble">;
     options: ChartConfiguration<"bubble">["options"];
   } | null>(null);
 
+  console.log("categorizedDataPoints", categorizedDataPoints);
```
pierluca, line 222: forgotten log?

frontend/src/components/dashboard/Analysis.tsx
```diff
@@ -364,7 +435,10 @@ const Analysis = ({
               }
               markAsHandled(evt.native);
 
-              setSplits((splits) => splits.filter((s) => s !== split));
+              dispatch({
+                type: AnalyzerStateActionType.setSplits,
+                splits: (splits) => splits.filter((s) => s !== split),
+              });
```
pierluca, line 441: This could be dispatched as a `removeSplit` and correspondingly `addSplit` below, making the logic testable exclusively within the dispatch.

frontend/src/components/dashboard/Analysis.tsx
Tyratox: Good catch - this makes it much cleaner! Somehow I have not noticed when writing, oops ^^

frontend/src/components/dashboard/CodeComparison.tsx
```diff
@@ -589,16 +634,21 @@ const CodeComparison = ({
...
-                          bookmarkedSolutionIds.filter(
-                            (id) => id !== rightDataPoint.analysis.solutionId,
-                          ),
-                        );
+                        dispatch({
+                          type: AnalyzerStateActionType.setBookmarkedSolutions,
+                          bookmarkedSolutionIds:
+                            state.bookmarkedSolutionIds.filter(
+                              (id) => id !== rightDataPoint.analysis.solutionId,
+                            ),
```
pierluca, line 642: I'm wondering if here we should do an `addBookmarkedSolution` / `removeBookmarkedSolution` instead, moving the logic to the reducer as well

frontend/src/components/dashboard/DissimilarityAnalysis.tsx
pierluca, line 1: Bound to change, not reviewing now.

pierluca (changes requested):

frontend/src/components/dashboard/Analyzer.state.ts
```diff
@@ -177,8 +235,21 @@ export const analyzerStateReducer = (
...
     case AnalyzerStateActionType.setSelectedSolution:
       return { ...state, selectedSolutionId: action.selectedSolutionId };
-    case AnalyzerStateActionType.setBookmarkedSolutions:
-      return { ...state, bookmarkedSolutionIds: action.bookmarkedSolutionIds };
+    case AnalyzerStateActionType.addBookmarkedSolution:
+      return {
+        ...state,
+        bookmarkedSolutionIds: [
+          ...state.bookmarkedSolutionIds,
+          action.solutionId,
```
pierluca, line 243: This can append multiple times the same ID. We should probably filter pre-existing IDs.

pierluca (approved):

PR #166, opened 2025-01-02
Various fixes requested by Laila & code reviews

pierluca:


frontend/src/components/dashboard/Analysis.tsx
```diff
@@ -119,6 +121,165 @@ const onLeaveLabel = (context: EventContext) => {
...
+        return byGroupKey;
+      },
+      {} as { [groupKey: string]: CategorizedDataPoint[] },
+    ),
+  ).map(([key, groupDataPoints]) => ({
+    key,
+    name: groupDataPoints[0].groupName,
+    analyses: groupDataPoints.flatMap((d) => d.analyses ?? []),
+  }));
+}
```
pierluca, line 281: All this code was extracted as-is from other places.
It's a first step in a broader refactoring.

frontend/src/components/dashboard/DissimilarityAnalysis.tsx
pierluca, line 1: This avoids a problem where N = 1 and k = 2, which doesn't work well for `n choose k` :-)

frontend/src/components/encryption/StudentName.tsx
```diff
@@ -124,7 +126,7 @@ export const StudentName = ({
   return (
     <NameWrapper>
       {name}
-      {isAnonymousUser
+      {isAnonymousUser && showActualName
         ? " (" + intl.formatMessage(messages.anonymous) + ")"
```
pierluca, line 130: no point in showing `(anonymous)` if we want the user to be anonymized :-)

frontend/src/utilities/navigation-observer.ts
Tyratox, line 1: Very cool :)

PR #172, opened 2025-01-03
[CRT-116] invalidate task file cache when changing file

pierluca:
`useSWR` was keeping an old file in its cache, even when a new file was posted to the server.
This PR mutates that cache to ensure that the new file is used.

### DIFF
--- a/frontend/src/api/collimator/hooks/tasks/useDeleteTask.ts
+++ b/frontend/src/api/collimator/hooks/tasks/useDeleteTask.ts
@@ -4,19 +4,19 @@ import { tasksControllerRemoveV0 } from "../../generated/endpoints/tasks/tasks";
 import { useRevalidateTaskList } from "./useRevalidateTaskList";
 import { useAuthenticationOptions } from "../authentication/useAuthenticationOptions";
 
-type DeleteUserType = (id: number) => Promise<DeletedTask>;
+type DeleteTaskType = (id: number) => Promise<DeletedTask>;
 
 const fetchAndTransform = (
   options: RequestInit,
   id: number,
-): ReturnType<DeleteUserType> =>
+): ReturnType<DeleteTaskType> =>
   tasksControllerRemoveV0(id, options).then(DeletedTask.fromDto);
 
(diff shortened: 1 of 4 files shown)


frontend/src/api/collimator/hooks/tasks/useUpdateTask.ts
```diff
@@ -56,10 +57,17 @@ const fetchAndTransformFile = (
...
   return useCallback(
     (id, updateTaskFileDto) =>
-      fetchAndTransformFile(authOptions, id, updateTaskFileDto),
-    [authOptions],
+      fetchAndTransformFile(authOptions, id, updateTaskFileDto).then(
+        (result) => {
+          revalidateTaskFile(id, updateTaskFileDto.file);
+
+          return result;
+        },
```
pierluca, line 69: this is the fix :)

frontend/src/api/collimator/hooks/tasks/useDeleteTask.ts
```diff
@@ -4,19 +4,19 @@ import { tasksControllerRemoveV0 } from "../../generated/endpoints/tasks/tasks";
 import { useRevalidateTaskList } from "./useRevalidateTaskList";
 import { useAuthenticationOptions } from "../authentication/useAuthenticationOptions";
 
-type DeleteUserType = (id: number) => Promise<DeletedTask>;
+type DeleteTaskType = (id: number) => Promise<DeletedTask>;
```
AthenaPK, line 7: Minor nitpick: unless this is already the convention in the project, I would avoid using `type` in the name, since this information is included in the type.

```suggestion
type DeleteTask = (id: number) => Promise<DeletedTask>;
```

AthenaPK (approved):
LGTM 🚀

frontend/src/api/collimator/hooks/tasks/useDeleteTask.ts
pierluca: Good point, but copied over from some other files where indeed it's more or less the convention.
I would also be wary of having something called `DeleteTask` as it seems more of an action than a  thing.

PR #179, opened 2025-01-07
[CRT-85] Enable selection of solutions on chart

Tyratox:
image


frontend/src/components/dashboard/Analysis.tsx
```diff
@@ -47,14 +47,16 @@ import {
...
 
 type AdditionalChartData = {
   groups: {
     key: string;
     name: string;
     analyses: CurrentAnalysis[];
   }[];
-  isSelected: boolean;
+  isSelectedForComparison: boolean;
+  isManuallySelected: boolean;
```
pierluca, line 59: They're both manually selected, I'm wondering if we could find better naming... but I have no ideas.

pierluca (approved):
LGTM, just a couple of minor improvement suggestions

pierluca (approved):
LGTM

PR #185, opened 2025-01-08
[CRT-43] Add E2E tests (and fix bugs discoverd through the tests)

Tyratox:


apps/scratch/src/components/customized-scratch-components/HashParserHOC.tsx
Tyratox, line 1: Copying this file was necessary since the scratch one always pushed to the history once..

backend/prisma/migrations/20250108080153_sorted_keys_for_fingerprint/migration.sql
Tyratox, line 1: Different implementations of the web crypto API export the JWK with keys in different orders. For the fingerprint to match, we need sorted keys.

backend/prisma/migrations/20250108080153_sorted_keys_for_fingerprint/migration.sql
```diff
@@ -0,0 +1,51 @@
...
+              'base64'
+            ),
+            '+',
+            '-'
+          ),
+          '/',
+          '_'
+        ),
+        '=',
+        ''
```
Tyratox, line 33: Base64 Url encoding

backend/prisma/migrations/20250108080153_sorted_keys_for_fingerprint/migration.sql
```diff
@@ -0,0 +1,51 @@
...
+      ),
+      E'\n',
```
Tyratox, line 35: Postgres automatically adds newlines - we don't want that

e2e/scripts/build-scratch.ts
```diff
@@ -0,0 +1,10 @@
+import { buildApp, CrtApp } from "../setup/helpers";
+
+const main = async (): Promise<void> => {
+  buildApp(CrtApp.scratch, "pipe", "pipe");
```
pierluca, line 4: what does "pipe", "pipe" do here? 😅

e2e/setup/helpers.ts
```diff
@@ -135,14 +150,30 @@ export const buildFrontend = (
...
     cwd: getFrontendPath(),
     shell: true,
     stdio: ["ignore", stdout, stderr],
   });
 };
 
+export const buildApp = (
+  app: CrtApp,
+  stdout: "pipe" | "ignore",
+  stderr: "pipe" | "ignore",
```
pierluca, line 164: since it seems we're using "pipe" everywhere, can we set it as default value? :-)

pierluca (changes requested):
I haven't gotten around to reviewing the tests themselves, but I think I've spotted a couple of issues.

e2e/setup/helpers.ts
Tyratox: I guess now we are so yes :)

PR #210, opened 2025-01-31
[CRT-25] Add sentry

Tyratox:


.github/workflows/tests.yml
```diff
@@ -145,6 +145,12 @@ jobs:
                     GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}  # Needed to get PR information, if any
                     SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
 
+            -   name: Upload backend source maps
```
pierluca, line 148: Why do this here and not in the deployment workflow? It might lead to undesirable side effects, where PRs change the source map but the deployed version is different, no?

apps/scratch/src/instrument.ts
```diff
@@ -0,0 +1,17 @@
...
+Sentry.init({
+  dsn: sentryDsn,
+  integrations: [
+    Sentry.browserTracingIntegration(),
+    Sentry.replayIntegration(),
+  ],
+  // Tracing
+  tracesSampleRate: 1.0, //  Capture 100% of the transactions
+  // Set 'tracePropagationTargets' to control for which URLs distributed tracing should be enabled
+  tracePropagationTargets: ["localhost"],
```
pierluca, line 13: This is wrong in production, isn't it?

apps/scratch/src/instrument.ts
```diff
@@ -0,0 +1,17 @@
...
+  // Session Replay
+  replaysSessionSampleRate: 0.1, // This sets the sample rate at 10%. You may want to change it to 100% while in development and then sample at a lower rate in production.
```
pierluca, line 15: The comment makes sense.
Can we have two configs, one for dev and one for prod?

backend/.gitignore
```diff
@@ -6,4 +6,6 @@
 /coverage
 /coverage-e2e
 /coverage-nest-e2e
-/.nyc_output
\ No newline at end of file
+/.nyc_output
+# Sentry Config File
+.sentryclirc
```
pierluca, line 11: We might want to put this in the global .gitignore

frontend/.babelrc.disabled
pierluca, line 1: Why disable?

frontend/.gitignore
```diff
@@ -32,4 +32,6 @@ next-env.d.ts
 
 *storybook.log
 
-.env.test
\ No newline at end of file
+.env.test
+# Sentry Config File
+.env.sentry-build-plugin
```
pierluca, line 37: Are you sure about this path?
Is it really a config file? 🤔

frontend/next.config.mjs
```diff
@@ -15,4 +16,49 @@ const nextConfig = {
   },
 };
 
+if (!process.env.BABEL_ENV || process.env.BABEL_ENV !== "coverage") {
+  // it seems that sentry and babel are not compatible but fortunately we need
```
pierluca, line 20: I'd be very very surprised if they weren't.
There's even this: https://www.npmjs.com/package/@sentry/babel-plugin-component-annotate

pierluca (changes requested):
Thanks for the good work!
A few points:
I reviewed from a phone, so I may have missed a couple of things and I definitely wasn't able to check others.
It seems as though there are multiple sentry projects - I'd try to keep everything in one project and enable distributed tracing such that we can correlate failures between frontend and backend, understand what action on the frontend triggered a backend problem, etc.
Same goes for communication across frontend & Scratch. By the way, can we instrument/log that into sentry as well?

pierluca:
Earlier, I missed that Prisma instrumentation was also left out.
We definitely want this:
https://docs.sentry.io/platforms/javascript/guides/express/configuration/integrations/prisma/

pierluca:
We may also want to instrument processes started to do the server-side generic AST transformation.

pierluca:
You may also want to handle graceful shutdown, if not otherwise taken care of by NestJS:
https://docs.sentry.io/platforms/javascript/guides/express/configuration/draining/

pierluca:
You might also want to enable trace propagation across front-end(+app) and back-end:
https://docs.sentry.io/platforms/javascript/guides/nestjs/tracing/trace-propagation/#how-to-use-distributed-tracing

apps/scratch/src/instrument.ts
Tyratox: I mean we can add as many environment variables as we like, yes. Do you see any other option you would like as a environment variable?

frontend/.gitignore
Tyratox: This was put there automatically, yes, but we don't need it :)

pierluca (approved):

PR #218, opened 2025-02-10
[CMT-30] Maintain freeze config when selecting sprite

Tyratox:


pierluca (approved):
Wow, good job spotting this one!

PR #226, opened 2025-02-10
[CMT-34] Allow the number of pairs to be 1

Tyratox:


pierluca (commented):
I set this to `2` as a quick (imperfect) fix because some other part of the codebase was crashing with values < 2. Are you sure this works now?

Tyratox:
Oh you are right!

pierluca (approved):

PR #247, opened 2025-02-27
[CRT-95] Add assertion names

Tyratox:


apps/scratch/src/hooks/useEmbeddedScratch.ts
```diff
@@ -91,26 +93,33 @@ export const useEmbeddedScratch = (
...
                 // once the project is backed up, run the project
                 vm.greenFlag();
               });
 
-              const { totalTests, passedTests } = await waitForAssertions;
+              const { passedAssertions, failedAssertions } =
+                await waitForAssertions;
+
+              const mapToTest = (assertion: Assertion): Test => ({
+                identifier: `${assertion.targetName} ${assertion.blockId}`,
```
Tyratox: @pierluca any opinion for the separator?

apps/scratch/src/extensions/assertions/index.ts
```diff
@@ -346,29 +355,51 @@ class AssertionExtension {
    * Callback for when the project assertions were run.
    */
   onProjectAssertionsRan = (): void => {
```
pierluca, line 357: This function seems incredibly convoluted now and not very readable, I'm not sure `reduce` here is the most natural way to approach the problem.

I asked GitHub Copilot to suggest an alternative, since I was struggling to phrase a better proposal myself. Here's the output - I haven't checked it for correctness w.r.t. to your implementation, but it *looks* significantly clearer to read and more intuitively understood.

```typescript
onProjectAssertionsRan = (): void => {
  const collectAssertions = (target) => {
    const state = getCustomState(target);
    return {
      passed: state.passedAssertions.map(a => ({ ...a, targetName: target.getName() })),
      failed: state.failedAssertions.map(a => ({ ...a, targetName: target.getName() })),
    };
  };

  const { passedAssertions, failedAssertions } = this.runtime.targets.reduce(
    (acc, target) => {
      const assertions = collectAssertions(target);
      return {
        passedAssertions: [...acc.passedAssertions, ...assertions.passed],
        failedAssertions: [...acc.failedAssertions, ...assertions.failed],
      };
    },
    { passedAssertions: [], failedAssertions: [] }
  );

  failedAssertions.forEach(assertion => this.runtime.emit("BLOCK_GLOW_ON", { id: assertion.blockId }));
  passedAssertions.forEach(assertion => this.runtime.emit("BLOCK_GLOW_OFF", { id: assertion.blockId }));

  this.runtime.emit("ASSERTIONS_CHECKED", passedAssertions, failedAssertions);
};
```

backend/prisma/migrations/20250226100923_add_test_table/migration.sql
```diff
@@ -0,0 +1,40 @@
+/*
+  Warnings:
+
+  - You are about to drop the column `passedTests` on the `Solution` table. All the data in the column will be lost.
+  - You are about to drop the column `totalTests` on the `Solution` table. All the data in the column will be lost.
+
+*/
```
pierluca, line 7: I think we can remove the warning 😅

backend/src/api/solutions/dto/solution-test.dto.ts
pierluca, line 1: I had understood identifier and context differently, but this makes perfect sense and it's even better, good job! 👍

pierluca (changes requested):
Lots of good things in here and some minor issues :-)

Be sure not to overuse `reduce()` without auxiliary functions, it really harms readability, especially after a long day 😅

apps/scratch/src/hooks/useEmbeddedScratch.ts
pierluca: Sorry I completely missed this message.

What characters are _never_ found in the targetName? `@`, `#`, `:`, `_` or something similar?

Tyratox: I think a slash should be fine but I could not find documentation on it :/

Tyratox: There is a target id but it is not exported and therefore probably not consistent. As for the block id, `jEk@4|i[#Fk?(8x)AV` is a valid id

Tyratox: Although there do not seem to be spaces or slashes meaning we can search from the end of the identifier finding the **last** occurrence and split there..

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/lastIndexOf

Tyratox: Found it: https://github.com/scratchfoundation/scratch-vm/blob/c563ee008b9d0e0ebe871b54e8dea4afad568717/src/util/uid.js#L11-L12. A slash is not find, a dollar sign is however!

backend/src/utilities/json-to-object-interceptor.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  beforeEach(() => {
+    interceptor = new (JsonToObjectsInterceptor(["field1", "field2"]))();
+    context = {
+      switchToHttp: jest.fn().mockReturnThis(),
+      getRequest: jest.fn().mockReturnValue({
+        body: {
+          field1: '{"key":"value"}',
+          field2: '{"key2":"value2"}',
+          field3: "not a json string",
```
pierluca, line 19: a string is valid JSON though 😅

backend/src/utilities/json-to-object-interceptor.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+  });
+
+  it("should parse JSON fields in the request body", () => {
+    interceptor.intercept(context, next);
+
+    const request = context.switchToHttp().getRequest();
+    expect(request.body.field1).toEqual({ key: "value" });
+    expect(request.body.field2).toEqual({ key2: "value2" });
+    expect(request.body.field3).toEqual("not a json string");
+    expect(request.body.field4).toEqual('{"not":"parsed"}');
```
pierluca, line 38: the difference here is subtle and I missed it at first.
You might want to add a comment to highlight the difference

```suggestion
    // parsed fields, expect JSON
    expect(request.body.field1).toEqual({ key: "value" });
    expect(request.body.field2).toEqual({ key2: "value2" });
    
    // unparsed fields, expect original value
    expect(request.body.field3).toEqual("not a json string");
    expect(request.body.field4).toEqual('{"not":"parsed"}');
```

if you set the values in a dictionary, you could more realistically do:
```suggestion
    // parsed fields, expect JSON
    expect(request.body.field1).toEqual(JSON.parse(body.field1));
    expect(request.body.field1).toEqual(JSON.parse(body.field2));
    
    // unparsed fields, expect original value
    expect(request.body.field1).toEqual(body.field3);
    expect(request.body.field1).toEqual(body.field4);
```

backend/prisma/sql/getCurrentAnalyses.sql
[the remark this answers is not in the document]
pierluca: This still seems buggy. The distinct is only on the studentId, not the test 🤔

apps/scratch/src/extensions/assertions/index.ts
pierluca: Much more readable, nice! 👍

pierluca (changes requested):
I believe there might still be a couple of issues, but much better 👍

Tyratox:
> This still seems buggy. The distinct is only on the studentId, not the test 🤔

I don't agree, this is intentional - let's discuss in person!

pierluca (approved):
LGTM!

PR #249, opened 2025-02-28
[CRT-118] Add time limit for scratch tasks

Tyratox:


apps/scratch/src/hooks/useEmbeddedScratch.ts
```diff
@@ -105,11 +107,18 @@ export const useEmbeddedScratch = (
...
+                      failedAssertions,
+                    }),
                 );
 
                 // once the project is backed up, run the project
                 vm.greenFlag();
+
+                setTimeout(() => {
+                  vm.stopAll();
+                }, maximumExecutionTimeInMs);
```
pierluca, line 121: Here we should check if the VM was still running. If so, we might want to pop up a message to the user saying something along the lines of "we stopped the run, it was taking too long..."

pierluca (commented):
Minor adjustment necessary for usability.

Also, there seems to be quite a bit of back-and-forth between seconds and milliseconds, since this is only a backstop and in Scratch it doesn't make sense to go into the millisecond range, should we only support seconds for now, store seconds, and only do `*1000` in the `setTimeout`? 🤔

pierluca (approved):

PR #250, opened 2025-02-28
[CRT-120] Switch to localStorage to support multiple tabs

Tyratox:
Note that this changes the threat model we can handle but it should be an acceptable useability-security trade-off.


pierluca (approved):
LGTM

pierluca (approved):
LGTM.

I hope we can collect coverage properly at some point though, I don't like Sonar always complaining about coverage :-(

PR #251, opened 2025-02-28
[CRT-71] Update frontend dependencies

Tyratox:


pierluca (approved):
Approving but the tests fail...

Tyratox:
That was a bit weird - the new playwright dependencies are seemingly incompatible with the old ones ^^

.github/workflows/tests.yml
```diff
@@ -49,6 +49,7 @@ jobs:
                 working-directory: ./backend
                 run: yarn install
 
+
```
pierluca, line 52: why this new whiteline? 😅 
Having read an interesting piece about linters being the tools to keep PRs free from such nitpicks, can we setup a linter to ban two consecutive newlines in all files? ^^

pierluca (approved):
LGTM!

.github/workflows/tests.yml
Tyratox: That would be nice, yes. Are you thinking of a project-wide linter for any file extension and if so, do you know one by any chance? :) The alternative would be to set up a YML linter for the .github folder which in itself might also be useful?

PR #264, opened 2025-03-21
[CRT-119] Adds support for reference solutions

Tyratox:
This PR adds support for reference solutions:
- in the data model and the backend
- the task editor
- the analysis chart
- the comparison dropdown

It does **not** change the behavior of the actual analysis yet

image
image
image


Tyratox:
Tests are to be added..

backend/src/api/tasks/tasks.service.ts
```diff
@@ -42,36 +100,180 @@ export class TasksService {
...
+          },
+        },
+      }),
+      // delete all solutions without any reference
+      this.prisma.solution.deleteMany({
+        where: {
+          referenceSolutions: { none: {} },
+          studentSolutions: { none: {} },
+        },
+      }),
```
pierluca, line 199: Do we really want to do this? It might be useful to keep for research.

backend/src/utilities/modify.ts
pierluca, line 1: Me likey 👍

backend/src/utilities/tuple-map.ts
```diff
@@ -0,0 +1,55 @@
+export class TupleMap<Key, Value> {
+  private readonly _map = new Map<string, Value>();
+
+  private getKey(key: Key): string {
+    return `${this.keyToString(key)}`;
+  }
+
+  constructor(private readonly keyToString: (key: Key) => string) {}
```
pierluca, line 8: You've documented everything except the specificity 😅
It might be worth mentioning what the expectations are on `keyToString`

frontend/src/api/collimator/models/class-properties.ts
```diff
@@ -3,4 +3,7 @@ type NonFunctionPropertyNames<T> = {
   [K in keyof T]: T[K] extends Function ? never : K;
 }[keyof T];
 
-export type ClassProperties<T> = Pick<T, NonFunctionPropertyNames<T>>;
+export type ClassProperties<T, TOmit extends string = ""> = Omit<
+  Pick<T, NonFunctionPropertyNames<T>>,
+  TOmit
+>;
```
pierluca, line 9: I haven't seen where this is used (swamped by the amount of changes), but it seems that adding an `Omit<,>` where needed would be more explicit and clear.

frontend/src/api/collimator/models/solutions/existing-student-solutions.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  static fromDto(dto: ExistingStudentSolutionDto): ExistingStudentSolution {
+    return new ExistingStudentSolution({
+      ...dto,
+      createdAt: new Date(dto.createdAt),
+      solution: ExistingSolution.fromDto(dto.solution),
+    });
+  }
+
+  // TODO: define what solution is to be displayed. e.g. the latest solution that solves the task
```
pierluca, line 37: this TODO is now addressed, right?

frontend/src/api/collimator/models/solutions/existing-student-solutions.ts
```diff
@@ -0,0 +1,51 @@
...
+  ): ExistingStudentSolution | null {
+    if (!solutions) {
+      return null;
+    }
+
+    const solutionsSortedByDate = solutions.toSorted(
+      (a, b) => b.createdAt.getTime() - a.createdAt.getTime(),
+    );
+
+    return solutionsSortedByDate[0];
```
pierluca, line 49: We might not need to sort the whole array since we only care about 1 result, `O(n)` is enough.

frontend/src/api/collimator/models/solutions/refrerence-analysis.ts
pierluca, line 1: typo in file name: refrerence

backend/prisma/migrations/20250319063941_add_reference_solutions/migration.sql
pierluca, line 1: Was this tested (extensively) to make sure we don't lose any data?
I can't spot any obvious issues but I'm mildly worried as the transformation is non-trivial.

pierluca (changes requested):

backend/prisma/migrations/20250319063941_add_reference_solutions/migration.sql
Tyratox: It was tested but let's double check together before deploying!

backend/src/api/solutions/solutions.service.ts
```diff
@@ -49,85 +94,218 @@ const latestAstVersion = AstVersion.v1;
...
+      analysis.solutionHash === null ||
+      analysis.testName === null ||
+      analysis.testPassed === null ||
+      analysis.genericAst === null ||
+      analysis.astVersion === null
+    ) {
+      throw new Error(
+        `Query response for 'getCurrentAnalyses' is missing reference analysis data. ${JSON.stringify(Object.keys(analysis))}`,
+      );
+    }
```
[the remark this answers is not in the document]
Tyratox: The data definition is in prisma.. But happy to put it in a separate static method in the service :)

backend/src/api/solutions/solutions.service.ts
```diff
@@ -49,85 +94,218 @@ const latestAstVersion = AstVersion.v1;
...
+        studentKeyPairId: analysis.studentKeyPairId,
+      });
+    }
+
+    return byAnalysisId;
+  }
+
+  private groupByReferenceAnalysis(
+    byAnalysisId: TupleMap<ReferenceKey, ReferenceAnalysis>,
+    analysis: getCurrentAnalyses.Result,
```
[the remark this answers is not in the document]
Tyratox: Well since the typing is slightly wrong we check for all the relevant fields in the group method. I think it reads somewhat nice now:
- input is some result
- method checks if it is of the expected type (which is documented in the return type)
- if is not of the expected type, an exception is thrown

frontend/src/api/collimator/models/solutions/existing-student-solutions.ts
pierluca: Fair enough, but since you usually love your `reduce` operations (😆), I was wondering why not just have a reduce with a single ternary expression. Wouldn't it be equally clear? 🤔

backend/src/api/solutions/solutions.service.ts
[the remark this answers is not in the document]
pierluca: How come the typing is wrong? 😢

backend/src/api/solutions/solutions.controller.ts
```diff
@@ -39,9 +40,19 @@ import {
...
-import { CurrentAnalysisDto } from "./dto/current-analysis.dto";
+import { CreateSolutionDto } from "./dto";
+import {
+  ExistingStudentSolutionDto,
+  StudentSolutionId,
+} from "./dto/existing-student-solution.dto";
+import { CurrentStudentAnalysisDto } from "./dto/current-student-analysis.dto";
+import { ReferenceAnalysisDto } from "./dto/reference-analysis.dto";
+import { CurrentAnalysesDto } from "./dto/current-analyses.dto";
+import { PatchStudentSolutionIsReferenceDto } from "./dto/patch-student-solution-is-reference.dto";
```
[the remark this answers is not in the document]
Tyratox: I tried configuring ESLint with

```
"import/no-internal-modules": [
      "error",
      {
        "allow": ["**/node_modules/**/*", "**/index.ts"]
      }
    ],
```

and it works on the frontend but I simply cannot get the same config running on the backend.. :(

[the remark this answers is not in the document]
Tyratox: https://classroomrt.atlassian.net/browse/CRT-123

pierluca (approved):
LGTM

PR #270, opened 2025-03-31
Fix coverage collection during e2e tests

Tyratox:


pierluca (approved):
Approved, but the scripts are starting to look a little too complicated for my taste, it might read like black magic to a new dev on the project.

I would consider adding a little bit of context to a README indicating which are the main entry points, and how the scripts are architected.

PR #292, opened 2025-04-22
[CRT-130] Allow student deanonymization

Tyratox:


frontend/src/hooks/useLocalStorage.ts
pierluca, line 1: This file would deserve its own tiny test suite.

pierluca (approved):
LGTM, but a couple of tests around the `useLocalStorage` hook would be nice

PR #293, opened 2025-04-22
[CRT-131] Reference grouping

Tyratox:


frontend/src/components/dashboard/hooks/automatic-grouping/reference-solution-clustering.ts
```diff
@@ -0,0 +1,86 @@
...
+    (analysis) => ({
+      reference: analysis,
+      analyses: [analysis],
+    }),
+  );
+
+  // for each non-reference solution, find the closest group and add it to that group
+  const groupAssignments = await Promise.all(
+    nonReferenceSolutions.map(async (analysis) => {
+      const distances = await Promise.all(
```
pierluca, line 57: similarly, `referenceDistances`

pierluca:
You probably went a little overboard with the renaming, but it's at least clearer 😅 
In hindsight, `assignment` confused me because it is very overloaded in an educational context, but Good Enough (TM) 👍

PR #294, opened 2025-04-22
[CRT-133] Maximize the minimum distance

Tyratox:
Note that there is also an important bugfix here: The .fill() resulted in each matrix row being the same array instance 🤦‍♂️

### DIFF
--- a/frontend/src/components/dashboard/hooks/useDissimilarAnalyses/index.ts
+++ b/frontend/src/components/dashboard/hooks/useDissimilarAnalyses/index.ts
@@ -120,7 +120,7 @@ export const maximizeDistanceSum = (
   getBestCombination(
     numberOfAnalyses,
     distances,
-    // the initial best distance is 0, since we want to maximize the minimum distance
+    // the initial best distance is 0, since we want to maximize the total distance
     0,
     // since we take sum the distance, we start with 0
     0,
--- a/frontend/src/components/dashboard/hooks/useDissimilarAnalyses/index.ts
+++ b/frontend/src/components/dashboard/hooks/useDissimilarAnalyses/index.ts
@@ -1,9 +1,11 @@
 import { useEffect, useState } from "react";
+import { match } from "ts-pattern";
 import { CurrentAnalysis } from "@/api/collimator/models/solutions/current-analysis";
 import { CurrentStudentAnalysis } from "@/api/collimator/models/solutions/current-student-analysis";
 import { getAstDistance } from "../ast-distance";
 import { DistanceType } from "../ast-distance/distance-type";
 import { choose, getCombinationsUnsafe } from "./combinatorics";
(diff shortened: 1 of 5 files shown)


frontend/src/components/dashboard/hooks/useDissimilarAnalyses/index.ts
```diff
@@ -69,33 +45,98 @@ const getDissimilarAnalyses = async <T extends CurrentAnalysis>(
...
+    Number.POSITIVE_INFINITY,
+    (d1, d2) => Math.min(d1, d2),
+    // the new distance is better if it is greater than the best distance (maximizing)
+    (bestDistance, newDistance) => bestDistance < newDistance,
+  );
+
+const getDissimilarAnalyses = async <T extends CurrentAnalysis>(
+  analysesIn: T[] | undefined,
+  numberOfAnalyses: number,
+  distanceType: DistanceType = DistanceType.pqGrams,
```
pierluca, line 107: I would consider the `DistanceMaximizationType` as an extra parameter, for now supporting only `DistanceMaximizationType.MinimumDistance`, though I'd argue we could trivially keep the old logic around for future reference.

pierluca (approved):
Minor remarks but LGTM overalls :-)

frontend/src/components/dashboard/hooks/useDissimilarAnalyses/index.ts
```diff
@@ -94,17 +104,36 @@ export const maximizeMinimumDistance = (
...
   );
 
+export const maximizeDistanceSum = (
+  numberOfAnalyses: number,
+  distances: number[][],
+): number[] =>
+  getBestCombination(
+    numberOfAnalyses,
+    distances,
+    // the initial best distance is 0, since we want to maximize the minimum distance
```
pierluca, line 123:
```suggestion
    // the initial best distance is 0, since we want to maximize the total distance
```

frontend/src/components/dashboard/hooks/useDissimilarAnalyses/index.ts
```diff
@@ -48,17 +50,25 @@ const computePairwiseDistance = async (
...
   numberOfAnalyses: number,
   distances: number[][],
-  initialDistance: number = 0,
-  combineDistances: (overallDistance: number, distance: number) => number = (
-    d1,
-    d2,
-  ) => d1 + d2,
+  initialBestDistance: number,
+  initialCombinationDistance: number,
+  combineDistances: (overallDistance: number, distance: number) => number,
```
pierluca, line 68: Don't change this now, but in the future we might want to take an array here - so that we can implement things like quadratic mean.

pierluca (approved):
Minor comment issue, but LGTM 👍

PR #348, opened 2025-08-20
Add storybook deployment

Tyratox:


frontend/package.json
```diff
@@ -31,7 +31,7 @@
     "test:storybook": "test-storybook --coverage --coverageDirectory coverage-storybook",
     "storybook": "cross-env NODE_OPTIONS=--preserve-symlinks storybook dev -p 6006",
     "storybook:await": "bash -c \"until curl --output /dev/null --silent --head --fail http://localhost:6006; do printf '.'; sleep 3; done\"",
-    "storybook:build": "storybook build",
+    "storybook:build": "storybook build && cd storybook-static && git init && git remote add origin git@github.com:crt25/collimator-storybook.git && git branch -M main && git add -A && git commit && git push -u origin github-pages/storybook --force",
```
pierluca, line 34: I'm not sure this is the right approach here. First of all, a bunch of these commands might fail when calling it multiple times. Secondly, we don't take advantage of the Git history. Thirdly, we can't build locally without automatically updating the storybook.

It might make sense to define a git submodule (not in the package.json), and in an action separate from the build, pull the latest version of the submodule, update it, and push it.

WDYT?

PR #360, opened 2025-09-16
[CRT-153] Anonymous session progress infinite loading indicator

bubudetp:
…e loading in anonymous sessions

### DIFF
--- a/frontend/src/components/dashboard/ProgressList.tsx
+++ b/frontend/src/components/dashboard/ProgressList.tsx
@@ -181,14 +181,16 @@ const ProgressList = ({
     isLoading: isLoadingSolutions,
   } = useAllSessionSolutions(classId, sessionId);
 
-  const studentList = useMemo(() => {
-    if (!klass || !solutions) return new Set<number>();
+  const studentIdArray = useMemo(() => {
+    if (!klass || !solutions) {
+      return new Array<number>();
+    }
 
     const studentIds = new Set([
       ...klass.students.map((student) => student.studentId),
       ...solutions.flatMap((s) => s.solutions.map((s) => s.studentId)),
     ]);
-    return studentIds;
+    return [...studentIds];
   }, [klass, solutions]);
 
   const progress = useMemo(() => {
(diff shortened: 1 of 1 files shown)


frontend/src/components/dashboard/ProgressList.tsx
```diff
@@ -181,17 +181,22 @@ const ProgressList = ({
     isLoading: isLoadingSolutions,
   } = useAllSessionSolutions(classId, sessionId);
 
-  const progress = useMemo(() => {
-    if (!klass || !session || !solutions) {
-      return [];
-    }
+  const studentList = useMemo(() => {
```
Tyratox, line 184: `studentIds` is more accurate I would say. You can rename the variable within the hook to sth like `studentIdSet`

frontend/src/components/dashboard/ProgressList.tsx
```diff
@@ -181,17 +181,22 @@ const ProgressList = ({
...
+    if (!klass || !solutions) return new Set<number>();
 
     const studentIds = new Set([
       ...klass.students.map((student) => student.studentId),
       ...solutions.flatMap((s) => s.solutions.map((s) => s.studentId)),
     ]);
+    return studentIds;
```
Tyratox, line 191: I think the rest will be easier to read if we already convert it to an array here:

```suggestion
    // convert the set to an array
    return [...studentIds];
```

(you will need to change the early return too)

Tyratox (changes requested):
Good work :)

I added a few minor comments.

I already gave you some feedback for the naming conventions. On top, you may want to ensure that the commit messages are not too long (i.e. so that everything fits).

On top, it is usually helpful to give a short high-level description in the PR of what the changes are supposed to achieve. No need to do it here since we discussed it together but just something for the future :)

frontend/src/components/dashboard/ProgressList.tsx
```diff
@@ -181,22 +181,26 @@ const ProgressList = ({
...
+    return [...studentIds];
   }, [klass, solutions]);
 
   const progress = useMemo(() => {
     if (!klass || !session || !solutions) {
       return [];
     }
 
-    const students = [...studentList].map((studentId) => {
+    const studentIds = studentIdArray;
```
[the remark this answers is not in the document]
bubudetp: Good catch, it feels redundant, I'll use the variable directly.

Tyratox (approved):

PR #365, opened 2025-09-23
[CRT-154] appended block to scratch stack

bubudetp:


Tyratox:
Some general feedback on top of the comments:

In general, pay attention to the git messages (max 72 characters). If you want to provide additional information to a commit, you can add a linebreak.

Regarding formatting changes: These problems may be solved once we fix the eslint plugin on your side.

For commit organization: The changes seem to be distributed in a non-obvious way across the commits. For a next time, splitting it into commits like:
1. Add new iframe-rpc message type 'x'
2. Handle incoming iframe-rpc message of type 'x' on the frontend
3. Send iframe-rpc messages of type 'x' from apps/scratch

If you want to more fine-grained that is fine too but now the change where you pass the `sendRequest` prop to the `Block` component is not in the same commit as the change where the `Block` component actually uses this new prop which makes it hard to review + hard to understand in the future if somebody looks at the git history. Analogously, `isBlockPartOfLargeStack` is introduced in one commit but not used in that commit at all.

For (automated) formatting changes it would be great to always have them in a separate commit because that is one we do not have to review. In general you had it several times that in a single commit message you say X + Y + Z. Just do three commits, one for X, one for Y and one for Z. Makes it easier to review and results in a better git history.

apps/scratch/src/containers/customized-scratch-containers/Blocks.tsx
```diff
@@ -72,30 +72,16 @@ import {
...
+import {
+  shouldTrackActivity,
+  trackStudentActivity,
+} from "../../utilities/student-activity-tracking";
 import ExtensionLibrary from "./ExtensionLibrary";
+import type { WorkspaceChangeEvent } from "../../types/scratch-workspace";
+import type { CrtContextValue } from "../../contexts/CrtContext";
+import type { StudentAction } from "../../utilities/student-activity-tracking";
 
 // reverse engineered from https://github.com/scratchfoundation/scratch-vm/blob/613399e9a9a333eef5c8fb5e846d5c8f4f9536c6/src/engine/blocks.js#L312
```
Tyratox, line 84: This comment should have been moved as well.

apps/scratch/src/containers/customized-scratch-containers/Blocks.tsx
```diff
@@ -72,24 +72,26 @@ import {
...
+  | "dragOutside"
+  | "endDrag"
+  | "delete"
+  | "var_create"
+  | "var_rename"
+  | "var_delete"
+  | "comment_create"
+  | "comment_change"
+  | "comment_move"
+  | "comment_delete";
```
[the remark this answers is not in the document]
bubudetp: Same as above ^^

frontend/src/components/Task.tsx
```diff
@@ -88,6 +88,22 @@ const Task = ({
...
 
+  const onStudentActivity = useCallback(
+    (action: string, blockId: string) => {
+      trackStudentActivity({
+        type: StudentActivityType.TASK_APP_ACTIVITY,
+        sessionId: session.id,
+        taskId: task.id,
+        appActivity: {
+          type: "SCRATCH_BLOCK",
+          data: `${action} block with id ${blockId}`,
```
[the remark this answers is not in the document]
bubudetp: Yes, I was so focused on just sending messages on append block that I forgot this part. I fixed that by storing the data in the backend as a JSON file instead of bytes and store more structured and meaningful data for analysis.

libraries/iframe-rpc/src/methods/post-student-activity.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export type PostStudentActivity = IframeRpcMethod<{
+  method: "postStudentActivity";
+  caller: RpcCaller.App;
+  parameters: {
+    action: Action;
+    blockId: string;
+  };
+  result: undefined;
+}>;
```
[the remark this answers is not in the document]
bubudetp: Yes, I adapted this part by replacing blockId by a record of data so it's adaptable for each application type to use.

apps/scratch/src/containers/customized-scratch-containers/Blocks.tsx
```diff
@@ -1093,26 +1093,31 @@ class Blocks extends React.Component<Props, State> {
...
+
+      if (!isBlockPartOfLargeStack(block, 2)) {
+        return;
+      }
+
+      try {
+        this.props.sendRequest?.("postStudentActivity", {
+          action: event.type,
+          blockId: event.blockId,
+        });
```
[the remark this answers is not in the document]
bubudetp: It should never be undefined so I changed the trackmoveparams interface to be stricter.

apps/scratch/src/containers/customized-scratch-containers/Blocks.tsx
```diff
@@ -1095,23 +1079,17 @@ class Blocks extends React.Component<Props, State> {
...
-      // if a block is moved or created, we should get this block and check the stacksize
-      // determining if all stacks are above the minimum required
-      if (this.props.canEditTask) {
-        return;
-      }
-
-      if (!event.blockId) {
-        return;
+    if (shouldTrackActivity(event, this.props.canEditTask)) {
+      if (["create", "move", "delete"].includes(event.type)) {
```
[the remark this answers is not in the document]
bubudetp: It makes sense to check if the action is still of a specific action type since we might no want to store student activities at the moment for other events than these. The && remark is completely valid, though I think a better  way is to add that check inside studentactivity which makes the code cleaner.

apps/scratch/src/utilities/student-activity-tracking.ts
```diff
@@ -22,6 +36,10 @@ export const trackStudentActivity = ({
   sendRequest,
   action,
 }: TrackMoveParams): void => {
+  if (!blockId || !action) {
```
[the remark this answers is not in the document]
bubudetp: Technically, we shouldn't accept those invalid parameters as they are part of core flow for the pipeline. It makes sense to strictly make sure they are present via the API.

apps/scratch/src/containers/customized-scratch-containers/Blocks.tsx
```diff
@@ -403,9 +406,9 @@ class Blocks extends React.Component<Props, State> {
...
       this._renderedToolboxXML !== nextProps.toolboxXML ||
       this.props.extensionLibraryVisible !==
-        nextProps.extensionLibraryVisible ||
+      nextProps.extensionLibraryVisible ||
       this.props.customProceduresVisible !==
-        nextProps.customProceduresVisible ||
+      nextProps.customProceduresVisible ||
       this.props.locale !== nextProps.locale ||
       this.props.anyModalVisible !== nextProps.anyModalVisible ||
       this.props.stageSize !== nextProps.stageSize
```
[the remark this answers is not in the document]
Tyratox: Were you now able to solve the problems with the extension or should we have a look together next time we meet in person? :)

apps/scratch/src/containers/customized-scratch-containers/Blocks.tsx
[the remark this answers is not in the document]
Tyratox: Where I was trying to with this is, does this other check belong here or should it be merged with the checks you perform within `shouldTrackActivity`?

apps/scratch/src/utilities/scratch-block-stack-utils.ts
```diff
@@ -1,29 +1,26 @@
...
-  const stacksize = getStackSize(current);
-  return stacksize >= minSize;
+
+  // Need to go top down to count the exact stack size
+  let size = 0;
+  let next: Block | null = current;
+  while (next) {
+    size++;
+    next = next.getNextBlock ? next.getNextBlock() : null;
+  }
```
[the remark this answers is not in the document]
Tyratox: Sorry my mind was not fully present when I wrote this comment because this will change the times `size++` is executed!

libraries/iframe-rpc/src/methods/post-student-activity.ts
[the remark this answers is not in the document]
Tyratox: That works, alternatively you could also use `unknown` as the type

apps/scratch/src/containers/customized-scratch-containers/Blocks.tsx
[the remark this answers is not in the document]
bubudetp: Not yet :(

I'd appreciate it if we can have a look at it together tomorrow!

apps/scratch/src/utilities/scratch-block-stack-utils.ts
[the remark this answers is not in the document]
bubudetp: No problem!

apps/scratch/src/containers/customized-scratch-containers/Blocks.tsx
```diff
@@ -1078,26 +1077,34 @@ class Blocks extends React.Component<Props, State> {
...
+    }
+
+    if (shouldTrackActivity(eventAction, event, this.props.canEditTask)) {
+      const block = this.getWorkspace().getBlockById(event.blockId ?? "");
 
+      if (!block) {
+        return;
+      }
+      (async () => {
+        const solution = await this.props.vm.saveProjectSb3();
```
Tyratox: I assume you mean this part here that does not have to be wrapped?

apps/scratch/src/utilities/student-activity-tracking.ts
```diff
@@ -1,61 +1,76 @@
...
+  sendRequest: CrtContextValue["sendRequest"];
+  action: StudentAction;
+  solution: Blob;
 }
 
-export type StudentAction = "create" | "move";
+export enum StudentAction {
+  CREATE = "create",
+  MOVE = "move",
+  DELETE = "delete",
```
Tyratox, line 15: Detail but the convention would to be to use `PascalCase` or `camelCase` for enums, not `ALLCAPS` :)

apps/scratch/src/utilities/student-activity-tracking.ts
```diff
@@ -1,61 +1,76 @@
...
-  return (
-    (event.type === "move" || event.type === "create") &&
-    event.recordUndo &&
-    !canEditTask &&
-    !!event.blockId
-  );
-}
+): boolean | undefined =>
+  (action === StudentAction.MOVE || action === StudentAction.CREATE) &&
+  event.recordUndo &&
```
Tyratox, line 34: I am still confused by this here and would advice documenting what the purpose of this check (and the other checks) is. Also the function is named `shouldTrackMove`. It may be worth renaming this to what we want to do, not what the underlying scratch data types are called.

apps/scratch/src/utilities/student-activity-tracking.ts
```diff
@@ -1,61 +1,76 @@
...
+  !!event.blockId;
 
-function shouldTrackMove(block: Block | null | undefined): boolean {
-  if (!block) {
-    return false;
-  }
-  return isBlockPartOfLargeStack(block);
-}
+const shouldTrackMove = (block: Block | null | undefined): boolean =>
+  !block ? false : isBlockPartOfLargeStack(block);
```
Tyratox, line 39: This is weird to read.

Better: `block ? isBlockPartOfLargeStack(block) : false`

and even better: `block && isBlockPartOfLargeStack(block)`

apps/scratch/src/utilities/student-activity-tracking.ts
```diff
@@ -1,61 +1,76 @@
...
 
-export type StudentAction = "create" | "move";
+export enum StudentAction {
+  CREATE = "create",
+  MOVE = "move",
+  DELETE = "delete",
+}
+
+const scratchToStudentAction: Record<string, StudentAction | null> = {
+  create: StudentAction.CREATE,
```
Tyratox, line 19: What is the create action? Is this related to the ticket i.e. appending blocks to existing stacks?

apps/scratch/src/utilities/student-activity-tracking.ts
```diff
@@ -1,61 +1,76 @@
...
-}
+const shouldTrackMove = (block: Block | null | undefined): boolean =>
+  !block ? false : isBlockPartOfLargeStack(block);
+
+const getBlockData = (block: Block): Record<string, unknown> => ({
+  blockId: block.id ?? "",
+  parentId: block.getParent?.()?.id ?? null,
+  position: block.getRelativeToSurfaceXY?.() ?? { x: 0, y: 0 },
+  childrenIds: block.getChildren?.().map((child) => child.id) ?? [],
+});
```
Tyratox, line 46: For this it may be worth having a quick discussion. It is a good starting point but I do not think the data like this will be useful for researchers. Moreover, this data will be event dependent, i.e. we want different information depending on the type.

For appends I think we for sure also need the type of block that is being appended.
I don't think the position will be useful honestly but I don't mind having it. However I do find it weird that it may be undefined and that we default to (0, 0). When can this be undefined? And if it actually can be undefined, I would either drop it altogether (if we decide we do not need it) or then also supply null to indicate the absence of this information instead of a placeholder value.

apps/scratch/src/utilities/student-activity-tracking.ts
```diff
@@ -1,61 +1,76 @@
...
-    return false;
-  }
-  return isBlockPartOfLargeStack(block);
-}
+const shouldTrackMove = (block: Block | null | undefined): boolean =>
+  !block ? false : isBlockPartOfLargeStack(block);
+
+const getBlockData = (block: Block): Record<string, unknown> => ({
+  blockId: block.id ?? "",
+  parentId: block.getParent?.()?.id ?? null,
```
Tyratox, line 43: I would decide on a value for "value is not present", i.e. `null`, and then use it consistently instead of placeholder values.

apps/scratch/types/scratch-blocks.d.ts
```diff
@@ -22,19 +22,16 @@ declare namespace ScratchBlocksExtended {
...
     inputList?: BlockInput[];
     isMonitored?: boolean;
 
     mutationToDom?: () => HTMLElement;
     domToMutation?: (element: HTMLElement) => void;
     getNextBlock?: () => Block | null;
     getParent?: () => Block | null;
+    getRelativeToSurfaceXY?: () => { x: number; y: number };
+    getChildren?: () => Block[];
+    id?: string;
```
Tyratox, line 34: Can these ever be undefined?

backend/prisma/schema.prisma
```diff
@@ -340,7 +340,7 @@ model StudentActivityApp {
   id Int @id
 
   type String
-  data Bytes
+  data Json
```
Tyratox, line 343: We'll need to quickly have a chat with @pierluca about this change.

backend/prisma/migrations/20251001090802_change_studentactivity_data_to_json/migration.sql
Tyratox, line 1: @AthenaPK will for sure have more discussions with you about commit messages but in general "Addressed comments" commits are not desired and here you also mix in new changes into the same commit. This certainly should have been a separate commit which would group the related changes together and allow the commit message to specify why it was done / necessary.

frontend/src/components/Task.tsx
```diff
@@ -89,16 +89,16 @@ const Task = ({
...
         sessionId: session.id,
         taskId: task.id,
         appActivity: {
-          type: "SCRATCH_BLOCK",
-          data: `${action} block with id ${blockId}`,
+          type: action,
+          data: data,
         },
-        solution: new Blob([], { type: "application/x.scratch.sb3" }), // dummy blob here, not used for user activity
+        solution: solution,
```
Tyratox, line 101: `solution: solution,` can be abbreviated as `solution,` (if the key and value are the same)

apps/scratch/src/utilities/student-activity-tracking.ts
bubudetp: Yeah, it wasn’t in the ticket, but I added it since creating blocks also impacts the student activity tracking. I thought it’d make sense to include for consistency, but I will remove it since it’s out of scope.

apps/scratch/src/types/scratch-student-activities/delete/index.ts
```diff
@@ -0,0 +1,8 @@
+export type {
+  StudentDeleteActivity,
+  DeleteActivityRequest,
+  DeleteStudentAction,
```
Tyratox: No need to import this :)

apps/scratch/src/types/scratch-student-activities/delete/types.ts
```diff
@@ -0,0 +1,24 @@
...
+  StudentAction,
+} from "../common";
+
+import {
+  DeletedBlockInfo,
+  DeletedBlockRecord,
+} from "../../../utilities/scratch-block";
+
+export interface StudentDeleteActivity extends BaseStudentAppActivity {
+  blockIdArray: DeletedBlockInfo | null;
```
Tyratox: If something is typed as an array, I would suggest omitting the array suffix, no need to include the type in the name. On the other hand, plural may make sense: `blockIds`.

But effectively these are not just block ids so maybe `deletedBlocks`?

apps/scratch/src/types/scratch-student-activities/common.ts
```diff
@@ -0,0 +1,41 @@
+import { Block } from "scratch-blocks";
+import { CrtContextValue } from "../../contexts/CrtContext";
+import { WorkspaceChangeEvent } from "../scratch-workspace";
+
+export interface BaseStudentAppActivity extends Record<string, unknown> {
+  blockId: string;
+  blockType: string;
+}
+
+export interface BaseActivityRequest {
```
Tyratox: No need to prefix this with `Base`

```suggestion
export interface ActivityRequest {
```

apps/scratch/src/utilities/scratch-student-activities/payloads/get-delete-payload.ts
```diff
@@ -0,0 +1,19 @@
...
+): StudentDeleteActivity | null => {
+  if (!event.blockId) {
+    return null;
+  }
+
+  return {
+    blockId: block.id,
+    blockType: block.type,
+    blockIdArray: block.blockIdArray,
+    sizeOfDeletion: block.blockIdArray.length,
```
Tyratox: `sizeOfDeletion` is redunant here :)

apps/scratch/src/types/scratch-student-activities/common.ts
```diff
@@ -0,0 +1,42 @@
...
+  solution: Blob;
+}
+
+export interface StudentActivityHandlerParams {
+  event: WorkspaceChangeEvent;
+  action: StudentAction;
+  canEditTask: boolean | undefined;
+  sendRequest: CrtContextValue["sendRequest"];
+  solution: Blob;
+  // For delete actions, the block may not be found in the workspace
```
Tyratox: If you use tsdoc, i.e. `/**`, the IDE will be able to display the information when you hover over it or use the key somewhere.

apps/scratch/src/types/scratch-student-activities/create/types.ts
```diff
@@ -0,0 +1,13 @@
+import {
+  BaseStudentAppActivity,
+  BaseActivityRequest,
+  StudentAction,
+} from "../common";
+
+export interface StudentCreateActivity extends BaseStudentAppActivity {
+  parentId: string | null;
```
Tyratox: I would also add a tsdoc saying that it is null if there is no parent / this is a top-level block (assuming that this is the case ^^)

apps/scratch/src/utilities/constants.ts
```diff
@@ -3,3 +3,4 @@
 export const sentryDsn = process.env.REACT_APP_SENTRY_DSN as unknown as string;
 
 export const defaultMaximumExecutionTimeInMs = 1000 * 30;
+export const ACTIVITY_TYPE = "SCRATCH" as const;
```
Tyratox: why use all caps here when all other constants use camelCase?

apps/scratch/src/utilities/scratch-block.ts
```diff
@@ -0,0 +1,80 @@
+import { WorkspaceChangeEvent } from "../types/scratch-workspace";
+import type { Block } from "scratch-blocks";
+
+const minStackSize = 2;
+
+export type DeletedBlockInfo = Array<{ id: string; type: string }>;
```
Tyratox: This will be easier to read:
```suggestion
export type DeletedBlockInfo = { id: string; type: string }[];
```

Also I would in general suggest to never define a type as `X[]` for some type X. Instead define the type `X` and use `X[]` wherever you want an array of that type.
Most of the time, you will need to use both and then it is much easier to deal with all use cases if you have the type of the array element available as well.

apps/scratch/src/utilities/scratch-block.ts
```diff
@@ -0,0 +1,80 @@
...
+export interface DeletedBlockRecord {
+  id: string;
+  type: string;
+  blockIdArray: DeletedBlockInfo;
+}
+
+function getStackSize(topBlock: Block): number {
```
Tyratox: I think I mentioned this before but wherever possible, prefer arrow functions.

apps/scratch/src/utilities/scratch-block.ts
```diff
@@ -0,0 +1,80 @@
...
+
+export const mapXmlBlockToBlock = (
+  xmlElement: Element,
+): DeletedBlockRecord | null => {
+  // Extract the block ID and type from the XML element
+  const blockId = xmlElement.getAttribute("id");
+  const blockType = xmlElement.getAttribute("type");
+
+  if (!blockId || !blockType) {
+    return null;
```
Tyratox: In general you return `null` in a lot of places and when reviewing it is unclear whether this are codepaths that are ever supposed to be taken. If so, then keep it but otherwise throw exceptions instead.

apps/scratch/src/utilities/scratch-block.ts
```diff
@@ -0,0 +1,80 @@
...
+
+      if (!id || !type) {
+        return null;
+      }
+
+      return { id, type };
+    })
+
+    // Filter out any null values in case of missing attributes
+    .filter((block): block is DeletedBlockInfo[number] => block !== null);
```
Tyratox: This is exactly what I was referring to when I said we should type the array elements and not the array ;)

apps/scratch/src/utilities/scratch-student-activities/pipeline.ts
```diff
@@ -0,0 +1,53 @@
...
+  trackMoveActivity,
+  trackDeleteActivity,
+} from "./index";
+
+type StudentActivityPipelineParams =
+  | CreatePipelineParams
+  | MovePipelineParams
+  | DeletePipelineParams;
+
+export const processStudentActivityPipeline = (
```
Tyratox: You should be able to get rid of this function completely. No need for another intermediate step here. This allows you to get rid of all the `pipeline` types and functions and `action` types.

Simply do this logic within `handleStudentActivityTracking`

apps/scratch/src/utilities/scratch-student-activities/pipeline.ts
```diff
@@ -0,0 +1,53 @@
...
+  params: StudentActivityPipelineParams,
+): void => {
+  switch (params.action) {
+    case StudentAction.Create:
+      trackCreateActivity({
+        block: params.block,
+        sendRequest: params.sendRequest,
+        solution: params.solution,
+        event: params.event,
+      });
```
Tyratox: I would suggest making all those functions directly accept `StudentActivityHandlerParams` and simply pass `params` here.

apps/scratch/src/utilities/scratch-student-activities/student-create.ts
```diff
@@ -0,0 +1,28 @@
...
+import { shouldTrackCreateBlock } from "./filters";
+import { getCreatePayload } from "./payloads";
+import { sendCreateActivity } from "./senders";
+
+export const trackCreateActivity = ({
+  block,
+  sendRequest,
+  solution,
+  event,
+}: BaseStudentAction): void => {
```
Tyratox:
```suggestion
}: StudentActivityHandlerParams): void => {
```

apps/scratch/src/utilities/scratch-student-activities/student-delete.ts
```diff
@@ -0,0 +1,29 @@
...
+import { shouldTrackDeleteBlock } from "./filters";
+import { getDeletePayload } from "./payloads";
+import { sendDeleteActivity } from "./senders";
+
+export const trackDeleteActivity = ({
+  block,
+  sendRequest,
+  solution,
+  event,
+}: DeleteStudentAction): void => {
```
Tyratox:
```suggestion
}: StudentActivityHandlerParams): void => {
```

apps/scratch/src/utilities/scratch-student-activities/senders/send-create-activity.ts
```diff
@@ -0,0 +1,21 @@
+import { ACTIVITY_TYPE } from "../../constants";
+import type {
+  CreateActivityRequest,
+  StudentCreateActivity,
+} from "../../../types/scratch-student-activities";
+
+export async function sendCreateActivity(
+  data: StudentCreateActivity,
+  { sendRequest, action, solution }: CreateActivityRequest,
```
Tyratox: I would drop the `CreateActivityRequest` type and either just accept `sendRequest` and `solution` as positional parameters OR define a new type shared between the `sendXActivity` functions bundling `sendRequest` and `solution` but **not** `action`. The `action` should be hardcoded in the respective function, otherwise you could call this function with action = "delete" which does not make sense.

apps/scratch/src/utilities/scratch-student-activities/senders/send-move-activity.ts
```diff
@@ -0,0 +1,21 @@
+import { ACTIVITY_TYPE } from "../../constants";
+import type {
+  MoveActivityRequest,
+  StudentMoveActivity,
+} from "../../../types/scratch-student-activities";
+
+export async function sendMoveActivity(
```
Tyratox: See `sendCreateActivity` comment

apps/scratch/src/utilities/student-activity-tracking.ts
```diff
@@ -0,0 +1,91 @@
...
+  type: string,
+): StudentAction | null => scratchToStudentAction[type] || null;
+
+export const shouldRecordStudentAction = (
+  action: StudentAction,
+  event: WorkspaceChangeEvent,
+  canEditTask: boolean | undefined,
+): boolean | undefined =>
+  // This condition ensures that we only track user-initiated block moves or creations
+  // when undo is enabled, the task is not editable, and the block has an ID.
```
Tyratox: When `undo` is enabled, the task is *not* editable? Let's have a chat about this one

backend/prisma/migrations/20251001090802_change_studentactivity_data_to_json/migration.sql
```diff
@@ -0,0 +1,9 @@
+/*
+  Warnings:
+
+  - Changed the type of `data` on the `StudentActivityApp` table. No cast exists, the column would be dropped and recreated, which cannot be done if there is data, since the column is required.
+
+*/
```
Tyratox: Replace this valid warning (and the warning from the bot) with an explanation why it is okay

backend/prisma/schema.prisma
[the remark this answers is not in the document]
Tyratox: The bot is correct here in theory, why thumbs down? Instead a reply with an explanation as to why it is fine would be better.

frontend/src/api/collimator/generated/models/trackAppStudentActivityDto.ts
```diff
@@ -10,5 +10,5 @@ export interface TrackAppStudentActivityDto {
   /** The application-internal type of the activity. */
   type: string;
   /** The binary app activity data encoded in base64. */
```
Tyratox: Update the comment

libraries/iframe-rpc/src/methods/post-student-activity.ts
```diff
@@ -0,0 +1,14 @@
+import { IframeRpcMethod } from "../remote-procedure-call";
+import { RpcCaller } from "../rpc-caller";
+
+export type PostStudentActivity = IframeRpcMethod<{
```
Tyratox:
```suggestion
export type PostStudentAppActivity = IframeRpcMethod<{
```

libraries/iframe-rpc/src/methods/post-student-activity.ts
```diff
@@ -0,0 +1,14 @@
...
+  method: "postStudentActivity";
```
Tyratox:
```suggestion
  method: "postStudentAppActivity";
```

libraries/iframe-rpc/src/methods/rpc-method-names.ts
```diff
@@ -0,0 +1,5 @@
+import { IframeRpcApplicationMessage, IframeRpcPlatformMessage } from ".";
+
+export type RpcMethodName =
```
Tyratox: I don't see this being used anywhere?

Tyratox (changes requested):
You made many things much much better but because there were so many changes again, there are a ton of comments again. Sorry ^^

libraries/iframe-rpc/src/methods/rpc-method-names.ts
bubudetp: I deleted this but it was used in tests

apps/scratch/src/types/scratch-student-activities/move/types.ts
```diff
@@ -1,7 +1,10 @@
 import { StudentAppActivity, ActivityRequest, StudentAction } from "../common";
 
 export interface StudentMoveActivity extends StudentAppActivity {
+  // The id of the block's previous parent (null if top-level)
```
Tyratox: thank you! next time use tsdoc and then it is perfect ;)

apps/scratch/src/types/scratch-blocks.ts
```diff
@@ -0,0 +1,19 @@
...
+  vm: VM;
+  canEditTask: boolean | undefined;
+  blocks: HTMLElement;
+  filterNonNull: typeof filterNonNull;
+  updateSingleBlockConfigButton: (
+    vm: VM,
+    blocks: HTMLElement,
+    opcode: string,
+    canEditTask: boolean | undefined,
+  ) => void;
```
Tyratox: This is fine here but I am still wondering what the thought was putting this into a separate file in a `types` directory rather than directly with where the function is defined that uses this as input.

Personally, I would have tended to put it there so that whenever the function changes I only need to touch that file but maybe I am missing something and there is a good reason for doing so :)

apps/scratch/src/utilities/scratch-block.ts
```diff
@@ -80,3 +83,81 @@ export const mapXmlBlockToBlock = (
...
+  eventAction,
+  vm,
+  canEditTask,
+  blocks,
+  filterNonNull,
+  updateSingleBlockConfigButton,
+}: BlockLifecycleParams): void => {
+  switch (eventAction) {
+    case StudentActionType.Create:
+    case StudentActionType.Delete:
```
Tyratox: Sorry for being a bit annoying here (and I missed it in the previous review): I am not sure `StudentActionType` fits well here and I am wondering whether we should generalize the type and make it not only a `StudentActionType` and rather a `WorkspaceEventType` or have both? What do you think?

apps/scratch/src/utilities/scratch-block.ts
```diff
@@ -80,3 +83,81 @@ export const mapXmlBlockToBlock = (
...
+  eventAction: StudentActionType,
+  canEditTask: boolean | undefined,
+): void => {
+  if (
+    eventAction === StudentActionType.Delete &&
+    event.recordUndo &&
+    event.blockId &&
+    canEditTask
+  ) {
+    // When switching sprites, blocks are deleted with recordUndo set to false
```
Tyratox: Did you intentionally move this comment?

backend/prisma/migrations/20251001090802_change_studentactivity_data_to_json/migration.sql
```diff
@@ -3,6 +3,10 @@
 
   - Changed the type of `data` on the `StudentActivityApp` table. No cast exists, the column would be dropped and recreated, which cannot be done if there is data, since the column is required.
 
+  Why this is okay:
+  The application is not yet in production, so there is no production data to worry about.
+  Furthermore, the application is still in active development, so any development data can be discarded. 
+
```
Tyratox: I actually disagree with those reasons but agree that we can do this.

The application is under active development, yes, but there is also a deployed version that is being used by the customer. We cannot simply delete their data, they will be super unhappy about that! (The migration is automatically run when we deploy).

The reason why it is okay is that so far the `StudentActivityApp` has not been used and therefore cannot contain any data! If you want to be extra sure, you could also convert the data but I would say there is no need given the table has not been used yet.

But we did always migrate the data for the development deployment so far as well. You can see this in the other migration files.

frontend/src/api/collimator/generated/models/trackAppStudentActivityDto.ts
```diff
@@ -9,6 +9,6 @@
 export interface TrackAppStudentActivityDto {
   /** The application-internal type of the activity. */
   type: string;
-  /** The binary app activity data encoded in base64. */
-  data: Record<string, unknown>;
+  /** The app activity data stored as a string. */
+  data: string;
```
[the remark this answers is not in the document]
bubudetp: Same comment as above.. This version of the file is outdated, but again it's my fault. I will watch out the commit history for the next time! sorry ^^

apps/scratch/src/utilities/scratch-block.ts
bubudetp: Yeah, you're right. We could either generalize it to WorkspaceEventType and then map the action later for the activity handler, or create a new BlockLifecycleEventType. While the second option can be done, we could simply reuse the workspacechange type.

PR #367, opened 2025-09-24
[CRT-150] Lock Cancel buttons during save in TaskModal

bubudetp:


bubudetp:
> The code changes make sense but I wonder whether this is a behavior we want. Did you discuss this with @pierluca by any chance?

What I understood from the ticket was to remove the ability to click on the cancel button after clicking on the save. Am I missing something?

From the description of the ticket: 
> It might make sense to disable the Cancel while Save is in progress.

Tyratox:
Sorry, you are correct and I missed that!

PR #370, opened 2025-10-01
Docs about iframe-rpc libraries

Aubrays:
In this documentation, I finally prefer to use embedded app and host app, because parent and child weren't very clear for me.

What do you think about this choice?

Which elements you think are missing?

### DIFF
--- a/docs/infrastructure/iframe-rpc.md
+++ b/docs/infrastructure/iframe-rpc.md
@@ -73,3 +73,59 @@ yarn demo
 
 ## React Integration: `iframe-rpc-react`
 
+### Overview
+
+`iframe-rpc-react` is build on top of `iframe-rpc` to provide React hooks that simplify integration.
+It encapsulates origin validation, lifecycle handling, and communication setup for a React environment.
+
+### Installation and Setup
+
+### Installation and Setup
+
+The library is installed with ClassMosaic parts.
+
+For development purpose on the library, you can install it via yarn:
+
+```sh
+# From: collimator/
+cd /libraries/iframe-rpc-react
+yarn install
+```
+
+### Hooks
+
+#### `useIframeParent`
+
+Used in the app (like Scratch)
+
+```ts
+const { isInIframe, hasLoaded, sendRequest } = useIframeParent(handleRequest);
+```
+
+##### Parameters
+
+- `handleRequest`: an object exposing RPC methods to the frontend
+
+##### Returns
+
+- `isInIframe`: `boolean` - if running inside an iframe.
+- `hasLoaded`: `boolean` - if the iframe has finished loading.
+- `sendRequest`: function - sends requests to the frontend.
+
+#### `useIframeChild`
+
+Used inside the frontend
+
+```ts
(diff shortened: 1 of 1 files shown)


docs/infrastructure/iframe-rpc.md
```diff
@@ -0,0 +1,134 @@
...
+-  the host application (aka the frontend), which embeds the iframe.
+-  the embedded application (like Scratch), which runs inside the iframe.
+
+It defines strict types for requests and responses, ensuring reliability and security.
+
+### Uses cases
+
+- Exchange data between the frontend ClassMosaic and an iframe hosting an app (like Scratch).
+- Call methods defined inside the iframe from the frontend.
+- Trigger events or actions in the iframe from the frontend.
```
Tyratox: It is a two-way communication protocol so method calls can also be initiated by the embedded app. Also this is exactly what happens for events; they are pushed to the host app from the embedded app.

docs/infrastructure/iframe-rpc.md
```diff
@@ -0,0 +1,134 @@
...
+
+The library is installed with ClassMosaic.
+
+Only for development purpose, you can install it via yarn:
+
+```sh
+# From: collimator/
+cd /libraries/iframe-rpc
+yarn install
+```
```
Tyratox: I think this may be a bit confusing. What was the intention of these two paragraphs?

docs/infrastructure/iframe-rpc.md
```diff
@@ -0,0 +1,134 @@
...
+### Installation and Setup
+
+The library is installed with ClassMosaic.
+
+Only for development purpose on the library, you can install it via yarn:
+
+```sh
+# From: collimator/
+cd /libraries/iframe-rpc-react
+yarn install
```
Tyratox: Same question as above

docs/infrastructure/iframe-rpc.md
```diff
@@ -0,0 +1,134 @@
...
+```
+
+##### Parameters
+
+- `handleRequest`: function handling incoming requests from the embedded app.
+- `onAppAvailable`: callback triggered when the app is ready.
+
+##### Returns
+
+- `sendRequest`: function - sends requests to the app.
```
Tyratox:
```suggestion
- `sendRequest`: function - sends requests to the embedded app.
```

Tyratox (changes requested):
I like the terminologies host & embedded app :)

docs/infrastructure/iframe-rpc.md
Aubrays: When I have tried the library for documentation, I would run the demo file and it doesn't work without installing the library as a standalone.

I suggest:

The library is installed by default with ClassMosaic.

Will run the demo? Install it via yarn:

```sh
# From: collimator/
cd /libraries/iframe-rpc
yarn install
```

docs/infrastructure/iframe-rpc.md
Aubrays: I suggest: 

### Uses cases

- Retrieve data between both applications
- Call methods in both directions for two-way interaction
- Trigger events or actions from one application to the other

PR #545, opened 2026-03-12
[CRT-382] Prevent connectOrCreate hash collision

bubudetp:
## Summary by cubic
Fixes unique constraint errors when updating tasks by pre-upserting `solution`s keyed by `(taskId, hash)` and connecting them, and blocks duplicate reference-solution files with clear, localized errors. Addresses CRT-382.

- **Bug Fixes**
  - Pre-upsert `solution` rows keyed by `(taskId, hash)` and use `connect` in nested `referenceSolution` create/update to avoid `connectOrCreate` collisions; add partial unique indexes (`WHERE deletedAt IS NULL`) for `ReferenceSolution(taskId, solutionHash)` and `StudentActivity(studentId, type, happenedAt)`.
  - Detect and reject duplicate reference-solution files by content hash; return `DUPLICATE_REFERENCE_SOLUTION` (400) from the controller; frontend uses `isApiErrorWithCode` to show a localized toast; unit tests cover duplicate and non-duplicate paths and verify successful updates. 
  - Standardize API errors: new backend `ApiError` base with `ErrorCode` (task errors extend it); frontend moves `ApiError`/`ConflictError` to `src/errors/api`; `fetch` forwards backend `message`/`errorCode`; i18n extraction includes `src/errors/**` with new localized messages.

<sup>Written for commit f47f68f969db4f0c7d92e727f3a2e986f8a20961. Summary will update on new commits.</sup>


backend/src/api/tasks/tasks.service.ts
```diff
@@ -20,6 +20,13 @@ export class TaskInUseByClassOrLessonWithStudentsError extends Error {
   }
 }
 
+export class DuplicateReferenceSolutionError extends Error {
+  constructor(message?: string) {
+    super(message);
+    this.name = "DuplicateReferenceSolutionError";
```
AthenaPK: Could we use the error code here instead of hardcoding the name? Or possibly remove this completely if we don't use the name for anything?

I know it's hardcoded in `TaskInOtherUsersLessonError` as well, but it probably shouldn't be (that one was my bad!).

AthenaPK (dismissed):
The fix looks correct to me, I'm only wondering if we might have the same issue in other places in the code?

Also, would it be possible to add some unit tests to verify the behaviour?

backend/src/api/tasks/tasks.service.spec.ts
```diff
@@ -23,4 +42,74 @@ describe("TasksService", () => {
...
+      });
+    });
+
+    it("should not throw DuplicateReferenceSolutionError when reference solutions have different file content", async () => {
+      const files = [
+        makeMulterFile(Buffer.from("content-a")),
+        makeMulterFile(Buffer.from("content-b")),
+      ];
+
+      await service.update(
```
pierluca, line 105: can we assert something here? This test has now no assertions.
The implementation of `service.update()` returning `Promise.resolve()` would pass this test

pierluca (approved):
Minor testing weakness, otherwise LGTM

PR #550, opened 2026-03-18
[CRT-356] Await kernel to be ready to execute custom commands

bubudetp:
## Summary by cubic
Blocks cell execution until Otter is installed and the student notebook is copied into the virtual FS with cwd set to `/student`. The kernel/session starts on panel open, Run All waits for readiness, and localized toasts show progress or errors to address CRT-356.

- **Bug Fixes**
  - Added a loading state that disables Run/Advance via settings and intercepts NotebookActions; shows toasts through a new `postMessage` RPC.
  - Started the kernel immediately; installs Otter, copies the notebook, and sets cwd to `/student`. A readiness promise resolves on success or rejects on failure; Run All awaits it and shows an error toast if install fails.
  - On init/install/locale-read failures, show localized error toasts and keep execution controls disabled.
  - Renamed `addKernelListeners` to `setupKernel`; moved `chdir` into `writeJsonToVirtualFilesystem(...)` and fixed Python indentation with a missing `Path` import.
  - Removed old toolbar overrides and manage button state programmatically.

- **New Features**
  - Localized toasts using `react-intl` with `@formatjs/cli`; added extract/fill/compile scripts and `en`/`fr` messages. Locale is read from settings with fallback and an error toast on read failure.
  - Shorter, closable app toasts and host-side handling of messages via the new `postMessage` RPC.

<sup>Written for commit 67f44a4c2ab8447a7c9f6dfe6af4c80b030ae6d7. Summary will update on new commits.</sup>


apps/jupyter/extensions/notebook-runner/src/commands/run-all-cells.ts
```diff
@@ -18,10 +19,15 @@ export const registerRunAllCellsCommand = (
...
 
+      console.log("awaiting for packages to be ready before running cells...");
+      await waitForPackagesReady();
+      console.log("packages are ready, running cells...");
+
       await NotebookActions.runAll(
         currentNotebook.content,
         currentNotebook.context.sessionContext,
       );
+      console.log("finished running cells");
```
pierluca:
```suggestion
      console.debug("awaiting for packages to be ready before running cells...");
      await waitForPackagesReady();
      console.debug("packages are ready, running cells...");

      await NotebookActions.runAll(
        currentNotebook.content,
        currentNotebook.context.sessionContext,
      );
      console.debug("finished running cells");
```

you probably want this to be debug logs :-)

apps/jupyter/extensions/notebook-runner/src/utils.ts
```diff
@@ -179,6 +179,7 @@ export const writeJsonToVirtualFilesystem = async (
   kernel: IKernelConnection,
   path: string,
   json: unknown,
+  chdir: string | null = null,
```
pierluca, line 190: The reason for this weird parameter should definitely be documented, as we're going to wonder why we did this a few months from now, and maybe be tempted to revert this 😅

apps/jupyter/extensions/notebook-runner/src/packages.ts
```diff
@@ -10,6 +10,21 @@ import {
...
+
+const resetPackagesReady = (): void => {
+  _packagesReady = new Promise<void>((resolve) => {
+    _setPackagesReady = resolve;
+  });
+};
+
+resetPackagesReady();
+
+export const waitForPackagesReady = (): Promise<void> => _packagesReady;
```
pierluca: This whole thing deserves some JSDoc-style documentation to explain why, what for, and how to use it.

I'm wondering though: do we still need this, considering the other initialization fix?

apps/jupyter/extensions/notebook-runner/src/utils.ts
bubudetp: Fairs 😅

apps/jupyter/extensions/notebook-runner/src/packages.ts
bubudetp: Do you mean resetting the promise on page load by `this`?

bubudetp:
Notification do not seem to be initialized at the point before initializing the kernel. To workaround that fact and still let users know, we agreed that adding a new rpc method to show a message to the user from the children to the parent iframe was necessary. 

The autoinstall packages flow seems to fail due to kernels being added after the install command is executed on task load. The kernels also don't really do anything until they're connected to the message flow. The issue seems to be coming from the timing, we're attempting to run the install command before the kernel is actually ready to process it. 

To fix this, we might need to try autoinstalling the packages straight from the beginning. This way we also wouldn't need the complex packages ready promise logic.

apps/jupyter/extensions/notebook-runner/src/packages.ts
pierluca: I literally meant about this part:
```
let _packagesReady: Promise<void>;
let _setPackagesReady: () => void;

const resetPackagesReady = (): void => {
  _packagesReady = new Promise<void>((resolve) => {
    _setPackagesReady = resolve;
  });
};

resetPackagesReady();

export const waitForPackagesReady = (): Promise<void> => _packagesReady;
```

whose usage is a little arcane , without further documentation 😅

apps/jupyter/extensions/notebook-runner/src/commands/run-all-cells.ts
```diff
@@ -22,6 +25,14 @@ export const registerRunAllCellsCommand = (
       console.debug(
         "awaiting for packages to be ready before running cells...",
       );
+
+      await sendMessage(
+        "Running Cells",
+        "Please wait while the required packages are being set up...",
```
pierluca: internationalization

apps/jupyter/extensions/notebook-runner/src/packages.ts
```diff
@@ -33,55 +55,70 @@ const autoInstallPackages =
...
+
+      if (isStudentNotebook) {
+        console.debug(
+          "Opened student notebook - copying notebook to the virtual filesystem to make tests available.",
+        );
+
+        let notebook: Contents.IModel | null = null;
+        try {
+          notebook = await contentsManager.get(notebookPath, { content: true });
+        } catch (error) {
```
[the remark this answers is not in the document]
bubudetp: great catch!

apps/jupyter/extensions/notebook-runner/src/packages.ts
```diff
@@ -54,47 +55,52 @@ const autoInstallPackages =
...
+        );
+        console.debug(
+          `Finished copying notebook to the virtual filesystem and changing working directory to ${studentWorkingDirectory}`,
+          kernel.id,
+        );
+      }
+
+      _setPackagesReady();
+    } catch (error) {
       _setPackagesReady();
```
pierluca: do we really want to _setPackagesReady()` if the above failed?

apps/jupyter/overrides.json
```diff
@@ -1,10 +1 @@
-{
-  "@jupyterlab/notebook-extension:panel": {
-    "toolbar": [
-      { "name": "restart", "disabled": true },
-      { "name": "restart-and-run", "disabled": true },
-      { "name": "save", "disabled": true },
-      { "name": "interrupt", "disabled": true }
```
pierluca, line 7: Actually we very much want these overrides, as they interact with/restart the kernel (supporting that creates a ton of work for us). `save` doesn't do what users expect it to do, that's why it's disabled too

apps/jupyter/extensions/notebook-runner/src/commands/run-all-cells.ts
```diff
@@ -23,7 +23,13 @@ export const registerRunAllCellsCommand = (
         "awaiting for packages to be ready before running cells...",
       );
 
-      await waitForPackagesReady();
+      try {
+        await waitForPackagesReady();
+      } catch (error) {
+        console.error("Cannot run cells, package installation failed:", error);
+        return;
```
pierluca, line 45: This is not visible to the user, so it becomes a silent failure.
How will the user know that something has gone wrong?

apps/jupyter/extensions/notebook-runner/src/index.ts
```diff
@@ -100,12 +100,14 @@ const plugin: JupyterFrontEndPlugin<void> = {
...
-    installPackagesWithLoadingState(
+    void installPackagesWithLoadingState(
       app,
       contentsManager,
       notebookTracker,
       loadingStateManager,
-    );
+    ).catch((error) => {
+      console.error("Failed to initialize package installation flow:", error);
+    });
```
pierluca: `void` here looks like a mistake.
Also, this function is async, we could `await` here.

If there's an exception, what can we (or the user) do?

pierluca (changes requested):

apps/jupyter/overrides.json
bubudetp: We disable them now through the loading state manager now

apps/jupyter/extensions/notebook-runner/src/commands/run-all-cells.ts
bubudetp: You're right we should notify them to reload the page.

pierluca (approved):

PR #552, opened 2026-03-19
[CRT-388][CRT-389] Block limits can be bypassed by dragging to other sprites & copy pasting

bubudetp:
## Summary by cubic
Prevents bypassing block limits via drag-to-sprite and copy/paste, and blocks deletion of task blocks in solve mode. Enforces `crtConfig.allowedBlocks` before VM events and undoes created/deleted blocks when limits or protections apply, addressing CRT-388, CRT-389, and CRT-392.

- **Bug Fixes**
  - Intercepts `create`, `endDrag`, and `delete` in `onWorkspaceChange` via `shouldPreventBlocksActions`; respects `canEditTask` and only forwards to `vm.blockListener` if allowed.
  - Looks up limits by opcode and checks counts with `countUsedBlocks(vm)`; undoes on `create` when over limit; undoes `delete` for task blocks.
  - Skips activity tracking on blocked actions; simplifies drag-to-sprite detection; adds types for `Workspace.getAllBlocks()` and `event.isOutside`.
  - Tests: remove initial-block-limit test, assert correct removal flow, and set `motion_movesteps` limit to 6 in `__tests__/tasks/assertion-task/crt.json`.

<sup>Written for commit be6ac0a5a4464c09c2e94de2dc90f0a4208695c6. Summary will update on new commits.</sup>

### DIFF
--- a/apps/scratch/src/blocks/helpers.ts
+++ b/apps/scratch/src/blocks/helpers.ts
@@ -84,7 +84,9 @@ export const wouldExceedLimits = (vm: VMExtended): boolean => {
     const hasLimit = typeof allowed === "number";
 
     if (hasLimit && count >= allowed) {
-      console.debug(`Block limit reached for ${opcode}: ${count} > ${allowed}`);
+      console.debug(
+        `Block limit reached for ${opcode}: ${count} >= ${allowed}`,
+      );
       return true;
     }
   }
--- a/apps/scratch/src/blocks/helpers.ts
+++ b/apps/scratch/src/blocks/helpers.ts
@@ -58,7 +58,6 @@ const findBlockInRuntime = (
 const shouldPreventBlockDeletion = (
   event: WorkspaceChangeEvent,
   vm: VMExtended,
-  workspace: ScratchBlocksExtended.Workspace,
   blockId: string,
 ): boolean => {
   if (event.type !== "delete") {
@@ -71,19 +70,18 @@ const shouldPreventBlockDeletion = (
   }
 
   if (block.isTaskBlock) {
-    workspace.undo(false);
     return true;
   }
 
   return false;
 };
 
-export const preventBlockActions = (
+export const shouldPreventBlocksActions = (
   event: WorkspaceChangeEvent,
   props: {
     canEditTask: boolean | undefined;
     vm: VMExtended;
-    workspace: ScratchBlocksExtended.Workspace;
+    workspace: ScratchBlocksExtended.Workspace | null;
     blockId: string | undefined;
   },
 ): boolean => {
@@ -93,11 +91,20 @@ export const preventBlockActions = (
     return false;
   }
 
+  if (!workspace) {
+    return false;
+  }
+
   if (shouldPreventBlockCreation(event, vm, workspace)) {
     return true;
   }
 
-  if (blockId && shouldPreventBlockDeletion(event, vm, workspace, blockId)) {
+  if (!blockId) {
+    return false;
+  }
+
+  if (shouldPreventBlockDeletion(event, vm, blockId)) {
+    workspace.undo(false);
     return true;
   }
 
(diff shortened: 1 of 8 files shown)


apps/scratch/src/blocks/helpers.ts
```diff
@@ -1,6 +1,93 @@
...
+        `Block ${opcode} is not allowed, blocking addition of blocks`,
+      );
+      return true;
+    }
+
+    // if allowed is a number, the block has a limit, it doesn't for every other type
+    const hasLimit = typeof allowed === "number";
+
+    if (hasLimit && count >= allowed) {
+      console.debug(`Block limit reached for ${opcode}: ${count} > ${allowed}`);
```
AthenaPK: Hmm the log is not consistent with the condition here.

apps/scratch/src/blocks/helpers.ts
```diff
@@ -1,6 +1,93 @@
...
+  if (canEditTask) {
+    return false;
+  }
+
+  // when a block is dragged outside the workspace onto a sprite thumbnail, scratch-blocks fires an endDrag event with the serialized block XML
+  // the vm uses this event to copy the blocks to the target sprite's workspace
+  const isBlockDraggedToSprite =
+    event.type === "endDrag" &&
+    // see EndBlockDrag in /scratch_blocks/core/scratch_events.js
+    "isOutside" in event &&
```
AthenaPK: Would it make sense to add `isOutside` and `xml` to `WorkspaceChangeEvent` instead of checking with `in` here?

AthenaPK (commented):
Overall, LGTM with a couple small suggestions : )
Also, would be nice to have a couple of unit tests for `shouldPreventBlockCreation` and `wouldExceedLimits` if time allows !

bubudetp:
Tests seem to fail here due to the changes introduced in this PR. The tests also fails locally. I was not able to reproduce the error by manually mimicking the failing test locally. The error seems to be coming from this line on the second iteration: https://github.com/crt25/collimator/blob/945f57f815d3ed4b45e9535e4badc7b06cbfe3fe/apps/scratch/src/__tests__/Solve.spec.tsx#L511 

The image below represents the failing test in debug mode with the browser console open.

image (6)

apps/scratch/src/blocks/helpers.ts
```diff
@@ -1,6 +1,96 @@
...
+    // see BlockDragger.prototype.endBlockDrag in /scratch_blocks/core/block_dragger.js
+    event.xml;
+
+  const isBlockCreated = event.type === "create";
+
+  if (!isBlockDraggedToSprite && !isBlockCreated) {
+    return false;
+  }
+
+  if (!wouldExceedLimits(vm)) {
```
pierluca: does the `event` include the new block? could we pass the block to `wouldExceedLimits` so as not to have to check against every opcode?

apps/scratch/types/scratch-blocks.d.ts
```diff
@@ -58,6 +58,7 @@ declare namespace ScratchBlocksExtended {
   class Workspace extends ScratchBlocks.Workspace {
     getFlyout(): Flyout | null;
     getBlockById(id: string): Block | null;
+    getAllBlocks(): Block[];
```
pierluca, line 61: is it me or this is not used?

pierluca (approved):
LGTM, just wondering if the implementation can be a tiny bit faster :)

apps/scratch/types/scratch-blocks.d.ts
bubudetp: Nope, it is used here: https://github.com/crt25/collimator/blob/bf57c635272f01c071fba9566af97f8870154e1d/apps/scratch/src/containers/customized-scratch-containers/Blocks.tsx#L630

PR #558, opened 2026-04-13
[CRT-392] Task blocks should not be deletable

bubudetp:
## Summary by cubic
Makes initial task blocks immutable in solve mode and prevents creating/duplicating blocks when limits would be exceeded. Blocked actions are immediately undone and skipped by the VM and activity tracking; also batches saving reference solutions to reduce DB work.

- **Bug Fixes**
  - Prevent deleting task blocks in solve mode (undo on delete).
  - Block creation/duplication when limits are exceeded (including drag-to-sprite); update tests to assert immutable initial blocks and correct limit updates for student-added blocks.

- **Refactors**
  - Replace per-file solution upserts with createMany + skipDuplicates when saving reference solutions.

<sup>Written for commit c23c95297d5f8197944f5d36e85334e13b2676b8. Summary will update on new commits.</sup>

### DIFF
--- a/apps/scratch/src/__tests__/Solve.spec.tsx
+++ b/apps/scratch/src/__tests__/Solve.spec.tsx
@@ -229,11 +229,9 @@ test.describe("/solve", () => {
     );
     await expect(moveSteps).toHaveText((moveStepsAllowedCount - 1).toString());
 
-    await page.removeAllNonFrozenBlocks();
+    const addedBlock = page.blocksOfCurrentTarget.last();
+    await page.removeBlock(addedBlock);
 
-    await expect(page.blocksOfCurrentTarget).toHaveCount(
-      task.blocksOfMainTarget - task.frozenBlocksOfMainTarget,
-    );
     await expect(moveSteps).toHaveText(
       getExpectedBlockConfigButtonLabel(task.crtConfig, "motion_movesteps"),
     );
@@ -419,26 +417,6 @@ test.describe("/solve", () => {
     );
 
     await expect(taskPage.blocksOfCurrentTarget).toHaveCount(0);
-  });
-
-  test("removing initial blocks does not increase the limit", async ({
-    page: pwPage,
-  }) => {
-    const { page, task } = await TestTaskPage.load(pwPage);
-
-    const { moveSteps, turnRight } = page.enabledBlockConfigButtons;
-
-    await expect(page.taskBlocks.catActor.editableBlock).toHaveCount(1);
-
-    await page.removeBlock(page.taskBlocks.catActor.editableBlock);
-
-    await expect(page.taskBlocks.catActor.editableBlock).toHaveCount(0);
-    await expect(moveSteps).toHaveText(
-      getExpectedBlockConfigButtonLabel(task.crtConfig, "motion_movesteps"),
-    );
-    await expect(turnRight).toHaveText(
-      getExpectedBlockConfigButtonLabel(task.crtConfig, "motion_turnright"),
-    );
   });
 
   test("loads assertions extension if task contains assertion blocks", async ({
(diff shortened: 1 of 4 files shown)


apps/scratch/src/containers/customized-scratch-containers/Blocks.tsx
```diff
@@ -1111,10 +1111,11 @@ class Blocks extends React.Component<Props, State> {
     }
 
     if (
-      shouldPreventBlockCreation(event, {
+      preventBlockActions(event, {
```
pierluca: rename suggestion: `should` hints at a `boolean` return type, whereas this hints at an imperative command (`prevent block actions!`). May I suggest `shouldPreventBlocksActions`?

apps/scratch/src/blocks/helpers.ts
```diff
@@ -50,6 +41,69 @@ export const shouldPreventBlockCreation = (
...
+
+const shouldPreventBlockDeletion = (
+  event: WorkspaceChangeEvent,
+  vm: VMExtended,
+  workspace: ScratchBlocksExtended.Workspace,
+  blockId: string,
+): boolean => {
+  if (event.type !== "delete") {
+    return false;
+  }
```
pierluca, line 65: what if the block is being moved to a different sprite? Should we block that too? 
Just thinking, not sure about impacts 🤔

apps/scratch/src/blocks/helpers.ts
```diff
@@ -50,6 +41,69 @@ export const shouldPreventBlockCreation = (
...
+
+  const block = findBlockInRuntime(vm, blockId);
+  if (!block) {
+    return false;
+  }
+
+  if (block.isTaskBlock) {
```
pierluca, line 72: Isn't this overly broad? 

As a teacher, I get to select what blocks are frozen, appendable or editable.

I see the point of ensuring that assertions are not removed, but this would imply that we can never remove any block from the original task 🤔 if the task is "debug and fix this", the fix might involve deleting something.

I've asked Engin & Laila for feedback on this one.

pierluca (commented):
I have more questions than answers after looking at the code. Corner cases I hadn't thought of. Waiting for Engin & Laila's answer.

Specification questions aside, implementation looks mostly good :)

apps/scratch/src/blocks/helpers.ts
pierluca: Feedback: this is just right! 👍

apps/scratch/src/blocks/helpers.ts
pierluca: Thinking some more, different sprite would just duplicate, not move.
All good here.

pierluca (approved):

apps/scratch/src/__tests__/Solve.spec.tsx
```diff
@@ -229,11 +229,9 @@ test.describe("/solve", () => {
     );
     await expect(moveSteps).toHaveText((moveStepsAllowedCount - 1).toString());
 
-    await page.removeAllNonFrozenBlocks();
+    const addedBlock = page.blocksOfCurrentTarget.last();
+    await page.removeBlock(addedBlock);
 
-    await expect(page.blocksOfCurrentTarget).toHaveCount(
-      task.blocksOfMainTarget - task.frozenBlocksOfMainTarget,
-    );
```
pierluca, line 236: would it make sense to check the following?
```
    await expect(page.blocksOfCurrentTarget).toHaveCount(
      task.blocksOfMainTarget - 1,
    );
```
to ensure the command above operated correctly?

bubudetp: isn't it `task.blocksOfMainTarget + 1` since removing blocks should increase the count?

PR #568, opened 2026-05-13
[CRT-402] Open latest notebook on task edit

bubudetp:
## Summary by cubic
Fixes CRT-402: Editing a Jupyter task now opens the latest student notebook (or the default notebook if none is provided), preserves the correct kernel, and blocks opening the teacher template (including via drag/drop). `loadTask` is now app-agnostic across Jupyter and Scratch.

- **Bug Fixes**
  - `EditTaskModal` sends `loadTask` with `{ task, language }` when `initialTask` exists; otherwise `{ useDefaultTask: true, language }`.
  - `iframe-rpc`: adds typed `LoadTask` params (`LoadTaskParamsWithTask` | `LoadTaskParamsWithoutTask`) and exports `isLoadTaskWithTask`.
  - Jupyter: when no task is provided or after `setLocale`, opens `notebookToOpen`; removed edit-mode template auto-open; intercepts document opens to redirect `task.ipynb` to `/student/task.ipynb`, blocking teacher template drag/drop.
  - Scratch: ignores `loadTask` without a task using `isLoadTaskWithTask`.

<sup>Written for commit 2d2cfd1bc85e5d778a3fa1ab3d1a1e29c491d650. Summary will update on new commits. <a href="https://cubic.dev/pr/crt25/collimator/pull/568?utm_source=github">Review in cubic</a></sup>


apps/jupyter/extensions/notebook-runner/src/iframe-api.ts
```diff
@@ -201,6 +201,13 @@ export class EmbeddedPythonCallbacks {
     try {
       await this.setJupyterLocale(request.params.language);
 
+      if (request.params.task === null) {
+        this.documentManager.openOrReveal(
+          EmbeddedPythonCallbacks.taskTemplateLocation,
+        );
+        return undefined;
+      }
```
pierluca: I like the idea but I don't like the `null` overload. We could make it explicit in the API. See the comment in `iframe-rpc`

libraries/iframe-rpc/src/methods/load-task.ts
pierluca, line 22: This will break `Scratch` if we don't support a nullable thing there.
More importantly, I don't think adding `null` here is desirable, as it could be interpreted in different ways.
Could we say that we have one of two things?

```suggestion
interface LoadTaskParamsBase {
  language: Language;
}

export interface LoadDefaultTaskParams extends LoadTaskParamsBase {
  task: Blob;
}

export interface LoadTaskParams extends LoadTaskParamsBase {
  useDefaultTask: true;
}

export type LoadTask = IframeRpcMethod<{
  method: "loadTask";
  caller: RpcCaller.CrtPlatform;
  parameters: LoadDefaultTaskParams | LoadTaskParams
  result: undefined;
}>;
```

In any case, don't forget to update the Scratch implementation!

apps/jupyter/extensions/notebook-runner/src/index.ts
```diff
@@ -155,21 +155,6 @@ const plugin: JupyterFrontEndPlugin<void> = {
...
-      if (mode === Mode.edit) {
-        try {
-          documentManager.openOrReveal(
-            EmbeddedPythonCallbacks.taskTemplateLocation,
-          );
-        } catch (error) {
-          console.error("Could not open template in edit mode:", error);
-        }
-      }
-    });
```
pierluca, line 179: Just wondering, during a reload (due to a locale change, for instance), do we still have a path to load the right task?

pierluca (changes requested):

frontend/src/components/modals/EditTaskModal.tsx
```diff
@@ -58,16 +58,20 @@ const EditTaskModal = ({
...
-      }
+      const locale = intl.locale as Language;
+      await executeAsyncWithToasts(
+        () =>
+          embeddedApp.sendRequest(
+            "loadTask",
+            initialTask
+              ? { task: initialTask, language: locale }
+              : {
+                  useDefaultTask: true,
```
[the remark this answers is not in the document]
pierluca: Much better, but indeed we still need to address Scratch support!

apps/jupyter/extensions/notebook-runner/src/index.ts
bubudetp: Yup

frontend/src/components/modals/EditTaskModal.tsx
```diff
@@ -58,18 +58,29 @@ const EditTaskModal = ({
       }
       wasInitialized.current = true;
 
-      if (initialTask) {
+      const language = intl.locale as Language;
+
+      if (taskType === TaskType.SCRATCH && initialTask) {
```
pierluca: we likely don't want the frontend to behave differently depending on the apps.
Let the apps decide how to handle the RPCs :)

```suggestion
      if (initialTask) {
```

apps/scratch/src/hooks/useEmbeddedScratch.ts
```diff
@@ -307,6 +314,12 @@ export class EmbeddedScratchCallbacks {
 
       console.debug(`${logModule} Loading project`);
 
+      if (!isLoadTaskWithTask(request.params)) {
+        throw new MissingTaskError(
+          this.intl.formatMessage(messages.missingTask),
+        );
```
pierluca: log and return instead of throw?

pierluca (changes requested):

pierluca (approved):

PR #569, opened 2026-05-13
[CRT-387][CRT-398][CRT-403][CRT-402] Fix soft-delete cascade + clean-up Makefiles

pierluca:
## Summary by cubic
Fix soft-delete cascade to be correct, idempotent, and fast. Adopt `mise` + Taskfiles repo-wide, speed up CI (incl. renamed API client check), align Playwright to 1.60, generate Swagger locally for `orval`, and pin Node 24 via `mise`. (Addresses CRT-398/CRT-403/CRT-402; tooling under CRT-387.)

- **Bug Fixes**
  - Cascade via relation filters with one `updateMany`/`deleteMany` per descendant; supports composite keys and flattens compound unique selectors.
  - Post-order traversal with `deletedAt: null` guards; idempotent across repeated `delete`/`deleteMany`; max-depth guard prevents cycles.
  - Fixed an invalid operation edge case with soft/hard relation mixes and pre-deleted parents.
  - Expanded e2e: three-level cascades, large `deleteMany` cascades, parent soft-delete doesn’t delete children, and isolation across trees.
  - Fixed Prisma include typings in Sessions service; jest-prisma now inherits the soft-delete extension for consistent test behavior.

- **Dependencies**
  - Repo-wide Taskfiles replace Makefiles; moved to `mise` (added `mise.toml`/lock); merged Jupyter Taskfile; symlink script is idempotent; docs updated; ignore `.task/`.
  - CI uses `jdx/mise-action` and `task`; renamed workflow to “API Client Generation Check”; faster `unit-tests` job; Postgres pinned to `postgres:18` (CI and `docker-compose`); backend tests require `DATABASE_URL`.
  - Backend can generate Swagger via `api:generate` (`--generate-swagger-only`) to `backend/docs/api.json`; frontend `orval` now reads that file.
  - Aligned `@playwright/test`/`playwright` to `^1.60.0` across `frontend`, `apps/scratch`, and `e2e`.
  - Frontend/Backend: split `lint`/`lint:fix`; added Storybook start/stop scripts; default backend `DATABASE_URL` updated; Node pinned to `24.x` via `mise`.
  - Shallow selected git submodules; excluded ANTLR grammars from Sonar and e2e coverage; deployment exports Sentry DSNs for frontend and scratch.

<sup>Written for commit 247c20fe1c3b012d50a1b46161e7b382ca35f760. Summary will update on new commits. <a href="https://cubic.dev/pr/crt25/collimator/pull/569?utm_source=github">Review in cubic</a></sup>


sonar-project.properties
```diff
@@ -14,4 +14,4 @@ sonar.cpd.exclusions=**/__tests__/**/*,**/*.spec.ts,**/*.spec.tsx,**/e2e/tests/*
 # Certain files are fully excluded from the analysis. This includes
 # - all generated files
 # - all coverage files
-sonar.exclusions=**/generated/**/*,./backend/coverage/**/*,./backend/coverage-nest-e2e/**/*,./backend/coverage-e2e/**/*,./frontend/coverage/**/*,./frontend/coverage-pw/**/*,./frontend/coverage-ct/**/*,./frontend/coverage-storybook/**/*,./apps/scratch/coverage/**/*
+sonar.exclusions=**/generated/**/*,**/backend/src/ast/antlr-grammars/**/*,./backend/coverage/**/*,./backend/coverage-nest-e2e/**/*,./backend/coverage-e2e/**/*,./frontend/coverage/**/*,./frontend/coverage-pw/**/*,./frontend/coverage-ct/**/*,./frontend/coverage-storybook/**/*,./apps/scratch/coverage/**/*
```
AthenaPK, line 17: Good catch! 👌

.github/workflows/deployment.yml
```diff
@@ -7,12 +7,13 @@ on:
   push:
     branches:
       - main
+      - bugfix/cascade-delete
```
AthenaPK: As discussed, this should probably be removed.

.github/workflows/tests.yml
AthenaPK: If we're pinning the version, we should probably do the same here and in the `orval-check.yml` workflow.

.github/workflows/deployment.yml
```diff
@@ -76,41 +78,16 @@ jobs:
...
-        working-directory: ./apps/jupyter
-        run: |
-          make install-deps
-          make build-otter
-          make init-extensions
-
-      - name: Install backend dependencies
-        working-directory: ./backend
-        run: yarn install
+
```
AthenaPK, line 77:
```suggestion
```

Taskfile.yml
```diff
@@ -0,0 +1,602 @@
...
+      - yarn test:cov
+
+  frontend:i18n:update:
+    desc: Update frontend i18n strings
+    dir: ./frontend
+    deps: [frontend:setup]
+    cmds:
+      - yarn i18n:update
+
+  frontend:sourcemap:upload:
```
AthenaPK, line 344: Should this be removed?

apps/jupyter/Taskfile.yml
```diff
@@ -0,0 +1,182 @@
...
+      - echo "No lint in Jupyter"
+
+  lint:fix:
+    deps: [lint]
+
+  test:
+    desc: Test the JupyterLite project
+    deps: [deps:install]
+    cmds:
+      - echo "No test in Jupyter"
```
AthenaPK: Shouldn't this call `task -d ./extensions test`? At least that is the behaviour of the deleted `Makefile`. I do think we might want to consolidate them in one as discussed though.

Taskfile.yml
```diff
@@ -0,0 +1,602 @@
...
+
+  apps:jupyter:setup:
+    desc: Install jupyter app dependencies
+    cmds:
+      - task -d apps/jupyter setup
+
+  apps:jupyter:build:
+    desc: Build jupyter app
+    cmds:
+      - task -d apps/jupyter build
```
AthenaPK, line 474:
```suggestion
      - task -d apps/jupyter build
      - task -d apps/jupyter/extensions build
```

mise.toml
```diff
@@ -0,0 +1,16 @@
+[tools]
+docker-cli = "29.4.0"
+node = { version = "24.15.0", postinstall = "corepack enable" }
+task = "3.50.0"
+opentofu = "1.11.6"
+python = "3.14"
+tflint = "0.62.0"
+
```
AthenaPK, line 8: Should we use mise for poetry too?

apps/jupyter/Taskfile.yml
```diff
@@ -0,0 +1,182 @@
...
+      - task: build:save-cache
+      - task: build:inject-iframe-script
+
+  lint:
+    desc: Lint the JupyterLite project
+    deps: [deps:install]
+    cmds:
+      - echo "No lint in Jupyter"
+
+  lint:fix:
```
AthenaPK, line 66: Should this be removed?

AthenaPK (changes requested):
Partial review, still a few commits to go through.

Feels like a good slashing of unnecessary complexity 👌 

As a general comment: we should probably update the documentation too, I can imagine that the developer setup guide could be simplified quite a bit!

mise.toml
pierluca: I really wanted to, but it doesn't work on Windows. Poetry is only supported on Linux/Darwin 😢

apps/jupyter/Taskfile.yml
pierluca: I actually added it on purpose. It's not yet there, but at least the behaviour is uniform across all apps / parts of the software system.

Taskfile.yml
pierluca: I actually lost a lot of time figuring out why that wasn't there. It might not be the most natural place to document this though 😅 what do you think?

sonar-project.properties
pierluca: So much time wasted debugging the wrong problem... this caused A TON of errors in the linter, I just added it here by extension 😅

apps/jupyter/extensions/notebook-runner/scripts/create-symlink.js
```diff
@@ -4,11 +4,34 @@ const path = require("path");
...
-  } else {
-    console.debug(`Symlink created: ${linkPath} -> ${target}`);
-  }
-});
+const createSymlink = () => {
+  fs.symlinkSync(target, linkPath, "dir");
+  console.debug(`Symlink created: ${linkPath} -> ${target}`);
+};
+
+const existingStats = fs.existsSync(linkPath) ? fs.lstatSync(linkPath) : null;
```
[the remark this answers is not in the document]
pierluca: Decided not to implement this, in light of how unlikely this is.

apps/jupyter/Taskfile.yml
```diff
@@ -0,0 +1,182 @@
...
+      - git submodule update --init --recursive
+
+  deps:install:
+    desc: Install Python dependencies via poetry
+    sources:
+      - pyproject.toml
+      - poetry.lock
+    cmds:
+      - pip install pipx
+      - pipx install poetry
```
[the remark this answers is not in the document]
pierluca: It doesn't fail for me.

backend/package.json
```diff
@@ -39,7 +40,7 @@
     "antlr:generate:python:parser": "java -cp ./antlr/antlr-4.13.2-complete.jar org.antlr.v4.Tool -Dlanguage=TypeScript ./src/ast/antlr-grammars/python/python3_13/PythonLexer.g4 ./src/ast/antlr-grammars/python/python3_13/PythonParser.g4 -o ./src/ast/converters/python/generated -visitor",
     "antlr:generate:python:copy": "shx cp ./src/ast/antlr-grammars/python/python3_13/TypeScript/* ./src/ast/converters/python/generated",
     "antlr:generate:python": "yarn antlr:generate:python:parser && yarn antlr:generate:python:copy",
-    "antlr:generate": "yarn antlr:download:grammers && yarn antlr:generate:python",
+    "antlr:generate": "yarn antlr:generate:python && yarn lint",
```
AthenaPK, line 43: Should this be `yarn lint:fix` ? Also, do we want to lint everything or only the generated files?

frontend/scripts/stop-storybook.mjs
```diff
@@ -0,0 +1,26 @@
+import { existsSync, readFileSync, unlinkSync } from "node:fs";
+import { spawnSync } from "node:child_process";
+
+const pidFile = "storybook.pid";
```
AthenaPK, line 4: Should we maybe use the constant from `start-storybook.mjs` here?

frontend/scripts/stop-storybook.mjs
```diff
@@ -0,0 +1,26 @@
...
+if (!existsSync(pidFile)) {
+  process.exit(0);
+}
+
+const pid = Number(readFileSync(pidFile, "utf8").trim());
+
+if (Number.isFinite(pid) && pid > 0) {
+  if (process.platform === "win32") {
+    spawnSync("taskkill", ["/F", "/T", "/PID", String(pid)], {
+      stdio: "ignore",
```
AthenaPK, line 15: Why `ignore` here?

backend/test/helpers/task/index.ts
```diff
@@ -137,6 +137,28 @@ export const createStudentSolution = async (
   });
 };
 
+export const createStudentSolutions = async (
```
AthenaPK, line 140: This doesn't seem to be used anywhere in the tests, should it be dropped?

apps/jupyter/extensions/notebook-runner/src/command.ts
```diff
@@ -10,7 +10,7 @@ export const runGradingCommand = "notebook-runner:run-grading";
 export const runAllCellsCommand = "notebook-runner:run-all-cells";
 
 export enum CommandType {
-  RunNotebook = "run_notebook",
+  RunNotebook =  "run_notebook",
```
AthenaPK, line 13: Typo?

.github/workflows/api-check.yml
```diff
@@ -34,83 +34,37 @@ jobs:
...
-            echo "The committed API client files do not match what orval generates."
+            echo "The committed API client files do not match what is generated."
             echo "This usually means the backend OpenAPI spec changed but the frontend"
             echo "API client was not regenerated."
             echo ""
             echo "To fix this:"
-            echo "  1. Start the backend: cd backend && yarn dev"
-            echo "  2. Regenerate the API client: cd frontend && yarn update:api"
+            echo "  1. Run: task frontend:api:generate"
             echo "  3. Commit the changes"
```
AthenaPK, line 68:
```suggestion
            echo "  2. Commit the changes"
```

.github/workflows/api-check.yml
```diff
@@ -34,83 +34,37 @@ jobs:
       BACKEND_API_URL: "http://localhost:3001/api-json"
 
     steps:
-      - name: Enable Corepack
-        run: corepack enable
-
       - name: Check out repository code
         uses: actions/checkout@v4
```
AthenaPK, line 38: Minor but since I'm seeing that the version in the other workflows got bumped, should we use the v6 checkout here too?

apps/jupyter/Taskfile.yml
```diff
@@ -0,0 +1,251 @@
...
+      - cmd: bash -c "if [ -f .cache/lab.html ] && [ -f {{.OUTPUT_DIR}}/lab/index.html ]; then cp .cache/lab.html {{.OUTPUT_DIR}}/lab/index.html; fi"
+        platforms: [linux, darwin]
+
+  build:save-cache:
+    internal: true
+    cmds:
+      - cp {{.OUTPUT_DIR}}/lab/index.html .cache/lab.html
+
+  build:inject-iframe-script:
+    cmds:
```
AthenaPK, line 243: This can probably be internal too.

apps/jupyter/Taskfile.yml
```diff
@@ -0,0 +1,251 @@
+version: '3'
+
+# helper to avoid repeating the dependencies of Jupyter extension files
+x-jupyter-ext-files: &jupyter_ext_files
```
AthenaPK, line 4: Interesting, thank you for opening my horizons about things one can do with yaml files : D

backend/src/prisma/extensions/soft-delete-extension.ts
AthenaPK, line 1: Thank you for renaming those files !👌 
(it was bothering me probably more than it should have : P)

backend/src/prisma/extensions/soft-delete-extension.ts
```diff
@@ -0,0 +1,296 @@
...
+                }
+
+                const existing = await txClient[clientModel].findUnique({
+                  where: args.where,
+                  include: args.include,
+                  select: args.select,
+                });
+
+                // Idempotent: deleting an already soft-deleted row returns it unchanged.
+                if (!!existing?.deletedAt) {
```
AthenaPK, line 213: Nice ! Just one question: if `deletedAt` is set to an object, e.g. ` deletedAt: { equals: null }`, is this going to be truthy?

pierluca: yes, but isn't the point of the extension that the implementation remains blissfully unaware of `deletedAt` ?

apps/jupyter/Taskfile.yml
pierluca: Happy to be of service. I hadn't used this kind of witchcraft in 10 years.
I hate that I had to (re-)figure this out, though...

apps/jupyter/Taskfile.yml
pierluca: Good point! Fixed in a subsequent PR.

.github/workflows/api-check.yml
pierluca: Good point! Fixed in a subsequent PR.

.github/workflows/api-check.yml
pierluca: Good point! Fixed in a subsequent PR.

PR #570, opened 2026-05-19
[CRT-399] Select kernel dialog pops up all the time

bubudetp:
Fix: preselect the kernel on load

---
## Summary by cubic
Preselects the default (Pyodide) kernel on notebook open and suppresses the Select Kernel dialog. Resolves CRT-399 so task creators can run cells without interruptions.

- **Bug Fixes**
  - Set `panel.sessionContext.kernelPreference.autoStartDefault = true` on `notebookTracker.widgetAdded` so the default kernel starts automatically and no prompt shows.

<sup>Written for commit 32d94a05fd4bb07f83eb1303dab8f4be695c2bc8. Summary will update on new commits. <a href="https://cubic.dev/pr/crt25/collimator/pull/570?utm_source=github">Review in cubic</a></sup>


apps/jupyter/extensions/notebook-runner/src/user-interface.ts
```diff
@@ -179,6 +180,44 @@ const redirectTaskFileOpensToStudentVersion = (
...
+
+  if (!specs) {
+    console.debug("No kernel specs found, cannot auto-select kernel");
+    return;
+  }
+
+  const pyodideKernelName = Object.keys(specs.kernelspecs).find(
+    (kernelName) =>
+      specs.kernelspecs[kernelName]?.display_name.includes("pyodide") ||
+      kernelName.includes("pyodide"),
```
pierluca: What are the names of the kernels (normal and grading one)? Do we have a more predictable/exact way of accessing the one we want?

The rest LGTM, I'm just worried about accidentally picking the wrong kernel.

pierluca: After discussion: only 1 kernel expected here. Make it an explicit expectation with a big loud error if this doesn't happen, then take that single one.

pierluca (approved):

apps/jupyter/extensions/notebook-runner/src/errors/task-errors.ts
pierluca, line 1: Is this still needed?

pierluca (changes requested):

pierluca (approved):

PR #573, opened 2026-05-22
[CRT-406] Switching a saved task from jupyter to scratch prevents scratch task creation

bubudetp:
## Summary by cubic
Fixes CRT-406: Switching between Jupyter and Scratch now only asks for confirmation when a task file exists (including files created in-session). This prevents the missing `project.json` error, avoids discarding the wrong file, and lets Scratch tasks be created and saved reliably.

- **Bug Fixes**
  - Guard type changes in `Select` via `shouldAllowValueChange`; show a “recreate task” modal in create mode and a “change type” modal in edit mode, keeping the current file/type until confirmed.
  - Prompt before replacing the task file when reference solutions exist and pass `clearAllReferenceSolutions` on save.
  - Rename `onBeforeChange` to `shouldAllowValueChange` in `Select` and `Checkbox`; apply to the `isPublic` toggle to confirm only when checking.
  - Add `testID` support to `Select` and `Checkbox`; use it in `UserForm` and `TaskForm`.
  - Alias `iframe-rpc` to a single path in `apps/scratch/next.config.ts` to prevent duplicate module instances when `resolve.symlinks=false`.

- **Tests**
  - Add comprehensive `TaskForm` UI tests (create/edit flows, in-session file creation, type-change confirmations, `isPublic` confirmation, save success/errors, disabled state).
  - Mock `EditTaskModal`, `Toaster`, and `navigation-observer`; remove `EmbeddedAppMock`.
  - Add Chakra UI jsdom polyfills and provider render helpers; load mocks in Jest setup, add `jest`/`node` types, and include an `ignore-console-errors` helper.
  - Fix e2e class deletion by introducing a `ClassDetailPageModel` and waiting for navigation before asserting list updates.

<sup>Written for commit 31bd88503a98f0b20516b2ecab207e60770e6105. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/crt25/collimator/pull/573?utm_source=github">Review in cubic</a>


frontend/src/components/form/Select.tsx
```diff
@@ -98,7 +101,13 @@ const InternalSelect = (
...
         name={name}
         value={value !== undefined ? [value.toString()] : []}
-        onValueChange={(v) => onValueChange?.(v.value[0])}
+        onValueChange={(v) => {
+          const newValue = v.value[0] as TaskType;
+          if (onBeforeChange && !onBeforeChange(newValue)) {
+            return;
+          }
+          onValueChange?.(newValue);
+        }}
```
pierluca, line 112: I have read this PR many times now, and I'm not sure I manage to understand it.
What is `onBeforeChange` supposed to do, what is the expected meaning of the boolean return value, and why would we prevent the notification of the value change?

frontend/src/components/task/TaskForm.tsx
```diff
@@ -281,10 +288,6 @@ const TaskForm = ({
   const cannotNavigate = useRef(false);
   const confirmedActionRef = useRef(false);
 
-  const [originalType, setOriginalType] = useState(
-    initialValues?.type ?? TaskType.SCRATCH,
-  );
```
pierluca, line 286: why do we want to remove the logic about preserving the original task type?

frontend/src/components/task/TaskForm.tsx
```diff
@@ -378,27 +381,22 @@ const TaskForm = ({
...
-    if (taskType === originalType) {
-      return;
-    }
-
-    if (hasTypeChanged) {
-      // at this point, if the user has already changed the type and confirmed,
-      // we can just allow switching without further confirmation
-
-      return;
-    }
```
pierluca, line 396: this used to be gated by a conditional such that, if we've already changed type once, we don't ask for confirmation every time. It seems we've lost the logic?

frontend/src/components/task/TaskForm.tsx
```diff
@@ -378,27 +381,22 @@ const TaskForm = ({
     onNavigate,
   });
 
-  useEffect(() => {
-    if (isTaskCreation) {
```
pierluca, line 382: it seems we're not taking this into account anymore?

pierluca (changes requested):
I'm a bit worried because I see a lot of logic changing, in a way I don't fully understand, for what should be a very narrow bug fix.
Do we need all of these changes?

frontend/src/components/task/TaskForm.tsx
bubudetp: This only existed because we wanted to know what value to revert the dropdown to when cancelling a type change with the useEffect approach. But with onBeforeChange we never commit the new type so there's nothing to revert. Keep an originalTarget to revert to is no longer needed. We keep hasTypechanged to decide whether to show the changeTaskFileConfirmation modal though when we are re editing an already changed task file.

frontend/src/components/form/Select.tsx
bubudetp: onBeforeChange is supposed to be between the value change before it's committed. It should return false to prevent the form field from updating and return true to allow it. 

in this case, we use it to show a confirmation modal when the user tries to change the task type while a task file already exists. The logic previously was in the useEffect where we'd detect a change, revert the select value with setValue and show the modal. 

The logic is now straightforward and easy to follow: select has the current value while the confirmation modal is open, and only if the user confirms we set it to the new type via setValue.

frontend/src/components/task/TaskForm.tsx
bubudetp: see first comment

frontend/src/components/task/TaskForm.tsx
bubudetp: see first comment

frontend/src/components/form/Select.tsx
pierluca: The naming is then clearly a problem. This is not a notification anymore, `on...` doesn't convey the purpose. How about `shouldAllowValueChange` or `validateValueChange` ?

frontend/src/components/task/TaskForm.tsx
pierluca: Ok, so does this mean that now if I change the task type, revert the task type, the field shows up as dirty and I can submit it and delete my original task without having brought any useful changes?

Phrased differently: do we still have a decent protection against users footgunning themselves? (I understand the dialog makes the risk explicit, I'm just wondering about what happens if a cat walks across the keyboard 😆)

bubudetp: Yeah here's the proof including all sorts of scenarios 😅 : 

https://jam.dev/c/03655dec-98bb-412a-a05b-0c765ceb85b0

frontend/src/components/form/Select.tsx
bubudetp: shouldAllowValueChange is indeed more appropriate!

pierluca (approved):

frontend/src/components/form/Select.tsx
```diff
@@ -44,6 +44,7 @@ type InternalSelectProps = {
   isDirty?: boolean;
   showEditedBadge?: boolean;
   errorMessage?: string;
+  "data-testid"?: string;
```
pierluca: there's a convention in React Native (Web) that the component prop is called `testID` (and rendered, just like you did, with a `data-testid` attribute), I would be tempted to adopt that because it'd be immediately obvious to any React (Native) developer :) 

What do you think?
I'm happy not to insist, but "data-testid" feels cumbersome to write 😅 

In this case (Select, Checkbox), I think the testID makes perfect sense, but I should also mention that the broader advice is not to use `data-testid` at all whenever possible: 
https://kentcdodds.com/blog/making-your-ui-tests-resilient-to-change
I don't think that really applies here though :-)

frontend/src/__tests__/mocks/components/EditTaskModal.tsx
pierluca, line 1: > Manual mocks are defined by writing a module in a __mocks__/ subdirectory immediately adjacent to the module. For example, to mock a module called user in the models directory, create a file called user.js and put it in the models/__mocks__ directory
(source: https://jestjs.io/docs/manual-mocks)

This would save you from having to mirror the whole hierarchy. 
You can then just do `jest.mock('...path.../EditTaskModal')` in your test :-)

frontend/src/__tests__/mocks/components/EditTaskModal.tsx
```diff
@@ -0,0 +1,42 @@
...
+          onSave({
+            file: new Blob(["task data"], { type: "application/json" }),
+            initialSolution: {
+              file: new Blob(["solution data"]),
+              failedTests: [],
+              passedTests: [
+                { identifier: "test1", name: "Test 1", contextName: null },
+              ],
+            },
+          });
```
pierluca, line 30: This mock in particular makes a whole lot of sense, since you're cutting down on the embedded app logic.

It might be useful to have a tiny comment explaining the purpose of the mock though, and how it cuts down on the complexity of EditTaskModal + TaskModal + Embedded App 😅

frontend/src/components/task/__tests__/jest/TaskForm.ui.spec.tsx
```diff
@@ -0,0 +1,807 @@
+/* eslint-disable @typescript-eslint/no-require-imports */
```
pierluca: it might make sense to put this only on the line where this happens, since there's only 1 require :-)

frontend/src/components/task/__tests__/jest/TaskForm.ui.spec.tsx
```diff
@@ -0,0 +1,807 @@
...
+// these exist in jsdom, so the real hook would crash.
+jest.mock("@/utilities/navigation-observer", () => ({
+  useNavigationObserver: jest.fn(() => jest.fn()),
+}));
+
+// The real toaster is a Chakra UI object whose .success/.error methods are not
+// Jest mock functions, so assertions like toHaveBeenCalledWith would throw. Which is why we mock it.
+jest.mock("@/components/Toaster", () => ({
+  toaster: { success: jest.fn(), error: jest.fn() },
+}));
```
pierluca: would it make sense to also provide a `__mocks__` folder with the mock implementations for these guys, so we don't need to do this manually here?

frontend/src/components/task/__tests__/jest/TaskForm.ui.spec.tsx
```diff
@@ -0,0 +1,807 @@
...
+
+// EditTaskModal embeds an external app in an iframe and communicates with it
+// over an RPC channel. The real modal would wait for RPC responses that
+// never arrive in jsdom. The mock bypasses the iframe entirely and calls
+// onSave directly with a hardcoded payload, simulating a completed edit.
+jest.mock("@/components/modals/EditTaskModal", () => ({
+  __esModule: true,
+  default: require("@/__tests__/mocks/components/EditTaskModal").default,
+}));
```
pierluca: as per the other comment, this could be reduced to a `jest.mock("@/components/modals/EditTaskModal")` by putting the mock in `@/components/modals/__mocks__/EditTaskModal` :-)

frontend/src/components/task/__tests__/jest/TaskForm.ui.spec.tsx
```diff
@@ -0,0 +1,807 @@
...
+// The real toaster is a Chakra UI object whose .success/.error methods are not
+// Jest mock functions, so assertions like toHaveBeenCalledWith would throw. Which is why we mock it.
+jest.mock("@/components/Toaster", () => ({
+  toaster: { success: jest.fn(), error: jest.fn() },
+}));
+
+// EditTaskModal embeds an external app in an iframe and communicates with it
+// over an RPC channel. The real modal would wait for RPC responses that
+// never arrive in jsdom. The mock bypasses the iframe entirely and calls
+// onSave directly with a hardcoded payload, simulating a completed edit.
```
pierluca: I'd basically copy over this into the `EditTaskModal` mock, as it's a perfect explanation for it 👍

frontend/src/components/task/__tests__/jest/TaskForm.ui.spec.tsx
pierluca, line 1: Is there any reason to not just put this in `frontend/src/components/task/TaskForm.spec.tsx` ? It's usually convenient to have them side-by-side, and `jest` in the path seems unnecessary.

I have seen that Tyratox left some tests with `__tests__/jest` , but I'm not sure if there's a good reason for that? 🤔

frontend/src/components/task/__tests__/jest/TaskForm.ui.spec.tsx
```diff
@@ -0,0 +1,807 @@
...
+
+      expect(screen.queryByTestId("isPublic")).not.toBeInTheDocument();
+    });
+  });
+
+  describe("Save with no task file", () => {
+    it("should show error toast and not call onSubmit", async () => {
+      const user = userEvent.setup();
+      const onSubmit = jest.fn().mockResolvedValue(null);
+      jest.spyOn(console, "error").mockImplementation(() => {});
```
pierluca: it seems this exact line is repeated in a bunch of places.
It might make sense to have a small test util function called `ignoreConsoleErrors()` :-)

frontend/src/components/task/__tests__/jest/TaskForm.ui.spec.tsx
```diff
@@ -0,0 +1,807 @@
...
+      expect(onSubmit).not.toHaveBeenCalled();
+    });
+  });
+
+  describe("Save errors", () => {
+    it("should show conflict error toast and call onConflictError when ConflictError is thrown", async () => {
+      const user = userEvent.setup();
+      const onConflictError = jest.fn();
+      const onSubmit = jest.fn().mockRejectedValue(new ConflictError());
+      jest.spyOn(console, "error").mockImplementation(() => {});
```
pierluca: since this is called everywhere in `describe("Save errors")`, you can put it in a `beforeEach()` structure, so as to avoid repetition.

frontend/src/components/task/__tests__/jest/TaskForm.ui.spec.tsx
pierluca, line 1: There are a lot of test IDs hardcoded in this file, repeated in many places.

It might make sense to have a micro "Page Object Model" (well, the equivalent for the component anyways) so as to be able to refactor the tests easily if the structure and/or test IDs and/or accessor change.

frontend/src/components/task/__tests__/jest/TaskForm.ui.spec.tsx
```diff
@@ -0,0 +1,807 @@
...
+      );
+
+      await user.click(screen.getByTestId("edit-task-button"));
+      await waitFor(() => {
+        expect(screen.getByTestId("mock-edit-task-modal")).toBeInTheDocument();
+      });
+      await user.click(screen.getByTestId("mock-save-task-button"));
+
+      await selectType(user, TaskType.JUPYTER);
+
```
pierluca, line 150: These tests, and perhaps others, rely on the precondition that `TaskType.SCRATCH` is selected by default. It might make sense to assert that precondition.

frontend/src/components/task/__tests__/jest/TaskForm.ui.spec.tsx
```diff
@@ -0,0 +1,807 @@
...
+      const isPublicCheckbox = screen.getByTestId("isPublic");
+
+      // Why we use not.toHaveAttribute("data-state", ...) instead of toBeChecked():
+      // data-testid is placed on ChakraCheckbox.Root (a <label>), not on
+      // HiddenInput (<input type="checkbox">). Putting it on HiddenInput does
+      // not work because Zag.js renders that input with `defaultChecked`.
+      //
+      // Zag.js always writes the correct controlled state to Root as a `data-state` attribute ("checked" / "unchecked"), regardless
+      // of what the uncontrolled DOM input holds. We therefore use not.toHaveAttribute("data-state", "checked") as the negation bypass which is the opposite of not.toBeChecked().
+      expect(isPublicCheckbox).not.toHaveAttribute("data-state", "checked");
```
pierluca: Interesting, but even more reason to have a "page object model"-kind of structure so that these idiosycransies can be fully captured there.

frontend/src/components/task/__tests__/jest/TaskForm.ui.spec.tsx
```diff
@@ -0,0 +1,807 @@
...
+        expect(toaster.error).toHaveBeenCalledWith(
+          expect.objectContaining({ id: "task-file-required" }),
+        );
+      });
+
+      expect(onSubmit).not.toHaveBeenCalled();
+    });
+  });
+
+  describe("Save errors", () => {
```
pierluca, line 688: it might make sense to have a boring "should not call onSubmit if not all fields are filled", albeit arguably this is arguably the most boring test you can think of :-)

pierluca (changes requested):
Very good, just a bunch of comments - hopefully easily addressed - to improve the overall maintainability.

frontend/src/components/task/__tests__/jest/TaskForm.ui.spec.tsx
bubudetp: yeah ^^

frontend/src/components/task/__tests__/jest/TaskForm.ui.spec.tsx
bubudetp: the jest.config.ts matches only **/__tests__/jest/**/*.spec.{ts,tsx} and the playwright-ct.config.ts matches only **/__tests__/playwright-ct/**/*.spec.tsx without this, jest would try to parse the playwright files and fail. We need to change the jest.config.ts and make sure to exclude the tests under the playwright folder. Should we do that?

frontend/src/__tests__/mocks/components/EditTaskModal.tsx
bubudetp: Thanks for sharing this !

frontend/src/components/form/Select.tsx
bubudetp: It makes sense to keep it simple though you're right 😅

frontend/src/components/task/__tests__/jest/TaskForm.ui.spec.tsx
pierluca: Ah, I see. Nah, don't bother, let's keep it as-is 👍

pierluca (approved):

PR #576, opened 2026-05-22
[CRT-391] UI does not show latest version of student code

bubudetp:
## Summary by cubic
Fixes CRT-391: the UI now always shows each student’s latest analyzed code across `StudentSolution` submissions and `StudentActivity`. Teachers can star/unstar activity-tracked analyses, and students no longer disappear while analysis runs.

- **Bug Fixes**
  - Current analyses now include activity-tracked work: add `getCurrentAnalysesWithActivities.sql` (+ soft-deleted), include `isStudentSolution` and nullable `studentSolutionId` in all queries, switch controller to the new query, and group by student+hash to keep multiple starred activity analyses separate; add tests for precedence/fallback, references, and grouping/merging.
  - Run analysis asynchronously on `StudentActivity.create`/`createMany` via `SolutionAnalysisService.performAnalysis` (`AstVersion.v1`); export from `SolutionsModule` and import in `StudentActivityModule`.
  - Support starring activity-tracked analyses: add `StudentActivity.isReference`, a PATCH endpoint that updates a specific activity by base64url `solutionHash` with auth (`canUpdateStudentActivityIsReference`, honors `includeSoftDelete`), and wrap the update in a DB transaction; regenerate frontend API; add `usePatchStudentActivityIsReference` and unify starring via `useStarAnalysis`; update `Showcase`, `StarAnalysisButton`, `StarSolutionButton`.
  - Propagate `isStudentSolution` across SQL/DTO/service/frontend; update `CurrentStudentAnalysis` (new flag, nullable `studentSolutionId`, stable `solutionId`), add and document `findAnalysisToDisplay`, and use it in the progress list and student page; allow `includeSoftDelete` in `canUpdateStudentSolutionIsReference`.

<sup>Written for commit 4184b10a447b38f25a4ba855b96dd26bb4076852. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/crt25/collimator/pull/576?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/src/api/solutions/solutions.service.ts
```diff
@@ -57,7 +57,7 @@ export type SolutionAnalysisCreateInput = Omit<
   "id"
 >;
 
-type StudentKey = [StudentId, TaskId, StudentSolutionId];
+type StudentKey = [StudentId, TaskId, StudentSolutionId | null];
```
pierluca: why do we need to allow null? This seems undesirable, as we'll have to check the typing everywhere we use this :-(

backend/src/api/student-activity/student-activity.service.ts
```diff
@@ -27,35 +30,49 @@ export class StudentActivityService {
...
-  createMany(
+  async createMany(
     student: Student,
     activityWithSolution: {
       activity: StudentActivityInput;
       solution: SolutionInput;
     }[],
   ): Promise<StudentActivity[]> {
-    const inputs: Prisma.StudentActivityCreateInput[] =
+    const results = await this.prisma.$transaction(
```
pierluca, line 43: this seems very wrong.
Having the `performAnalysis` in the transaction means that the analysis could start before the transaction is committed, or even if the transaction is rolled back.

backend/src/api/student-activity/student-activity.service.ts
```diff
@@ -27,35 +30,49 @@ export class StudentActivityService {
...
   ): Promise<StudentActivity> {
-    const input = this.buildActivityInput(student, activity, solution);
+    const result = await this.prisma.studentActivity.create({
+      data: this.buildActivityInput(student, activity, solution),
+      include: { solution: true },
+    });
+
+    // do not wait for the promise to resolve
+    // this will happen in the background
+    this.analysisService.performAnalysis(result.solution, latestAstVersion);
```
pierluca, line 73: why do we want to trigger this here? 🤔 
wasn't the analysis happening automatically in the background, before?

frontend/src/api/collimator/models/solutions/current-student-analysis.ts
```diff
@@ -30,7 +32,7 @@ export class CurrentStudentAnalysis extends CurrentAnalysis {
   }
 
   public override get solutionId(): string {
-    return `STUDENT:${this.studentSolutionId}`;
+    return `STUDENT:${this.studentSolutionId ?? this.solutionHash}`;
```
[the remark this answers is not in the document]
pierluca: I agree with the bot.

backend/src/api/solutions/solutions.service.ts
```diff
@@ -77,7 +77,7 @@ export type AnalysisWithoutId = {
 export type CurrentStudentAnalysis = AnalysisWithoutId & {
   studentId: number;
   sessionId: number;
-  studentSolutionId: StudentSolutionId;
+  studentSolutionId: StudentSolutionId | null;
```
pierluca, line 87: if there's no way around making this optional, it seems the proper way would be to allow `undefined` as a value

```suggestion
  studentSolutionId?: StudentSolutionId;
```

backend/src/api/solutions/solutions.service.ts
```diff
@@ -153,7 +153,7 @@ export class SolutionsService {
     const referenceAnalyses: getCurrentAnalyses.Result[] = [];
 
     for (const analysis of filteredAnalyses) {
-      if (analysis.studentSolutionId) {
+      if (analysis.studentId !== null) {
```
pierluca, line 164: wouldn't this be sufficient?

```suggestion
      if (analysis.studentId) {
```

backend/prisma/sql/getCurrentAnalyses.sql
pierluca, line 1: I would consider splitting this into two: one method that focuses exclusively on the "submitted" tasks (i.e. `getCurrentAnalyses.sql` unchanged), and another one that roughly follows your changes (e.g. `getCurrentAnalysesWithActivities`).

Higher level question: do we need this second method to merge both "submitted" solutions and activities? Or would activities include the submitted solutions anyways? 
In the latter case, we could just have a dedicated `getCurrentActivityAnalyses` that ignores the submitted solutions altogether (i.e. no `getCurrentAnalysesWithActivities`) 🤔 

If we do need to merge both, we should also be able to quickly generate a test for this, ensuring the two methods (getCurrentAnalyses, getCurrentAnalysesWithActivities), behave the same way when there are only "submitted" results, while returning different things if there's an extra `StudentActivity`.

pierluca (changes requested):
Higher order question: do these changes affect in any way the display of the task status in the dashboard? I.e. a task being started by a student, or completed.

frontend/src/api/collimator/models/solutions/current-student-analysis.ts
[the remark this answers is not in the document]
bubudetp: Makes sense

backend/src/api/student-activity/student-activity.service.ts
bubudetp: The analysis was never performing when creating student activities, it was for solutions though: https://github.com/crt25/collimator/blob/main/backend/src/api/solutions/solutions.service.ts#L566

backend/src/api/student-activity/student-activity.service.ts
bubudetp: performAnalysis is not called in the transaction though. If any transaction is rolled back, performAnalysis will never be called.

backend/src/api/solutions/solutions.service.ts
bubudetp: I reverted this change because and added a static type guard 'isSubmittedStudentAnalysis' instead

backend/src/api/solutions/solutions.service.ts
```diff
@@ -153,7 +169,7 @@ export class SolutionsService {
...
```
pierluca, line 164: just realized, won't this make student activities come out as reference solutions?
shouldn't we perhaps just use an `isStudent` field created at the DB level? 🤔

backend/src/api/solutions/dto/current-student-analysis.dto.ts
```diff
@@ -33,11 +33,12 @@ export class CurrentStudentAnalysisDto
...
     description:
-      "The student solutions's unique identifier, a positive integer.",
-    nullable: false,
+      "The student solutions's unique identifier. Null when the latest solution came from activity tracking rather than an explicit submission.",
+    nullable: true,
+    type: "number",
   })
   @Expose()
-  readonly studentSolutionId!: StudentSolutionId;
+  readonly studentSolutionId!: StudentSolutionId | null;
```
pierluca, line 41: do we use this field anywhere on the frontend? I'd consider removing this entirely, perhaps just transferring some bool like a `isReference` and/or `isStudentSolution`

frontend/src/components/dashboard/StarAnalysisButton.tsx
pierluca: how can I pick as a reference a student activity which has no `studentSolutionId` ?

pierluca (changes requested):

backend/src/api/solutions/dto/current-student-analysis.dto.ts
bubudetp: It's the key for the patch endpoint, if we remove it, it breaks the star/unstar flow entierly unless we re-key the backend endpoint and update the controller, service, auth check, frontend client, hook and star/showcase components. 

I don't think it's worth removing right now

backend/src/api/solutions/solutions.service.ts
```diff
@@ -397,6 +409,45 @@ export class SolutionsService {
...
+    if (!targetActivity) {
+      throw new NotFoundException(
+        `No analyzed student activity found for student ${studentId} in session ${sessionId} / task ${taskId} with the given solution hash`,
+      );
+    }
+
+    await this.prisma.studentActivity.update({
+      data: { isReference },
+      where: { id: targetActivity.id },
+    });
```
pierluca: this should probably happen in a transaction :)

frontend/src/api/collimator/models/solutions/current-student-analysis.ts
```diff
@@ -78,9 +72,9 @@ export class CurrentStudentAnalysis extends CurrentAnalysis {
         a instanceof CurrentStudentAnalysis && a.studentId === studentId,
     );
 
-    const starred = studentAnalyses.find((a) => a.isReferenceSolution);
     const current = studentAnalyses.find((a) => !a.isReferenceSolution);
+    const starred = studentAnalyses.find((a) => a.isReferenceSolution);
 
-    return starred ?? current ?? null;
+    return current ?? starred ?? null;
   }
```
pierluca, line 84: this is only correct if studentAnalyses are ordred by descending time.
1. it would be good to document this hypothesis
2. are they ordered? 😅

pierluca (changes requested):
Tiny fixes, but mostly LGTM 👍

frontend/src/api/collimator/models/solutions/current-student-analysis.ts
bubudetp: 1. 👍🏼 
2. It does not matter if they are ordered because the find calls works because there's at most one non reference entry to find, this is guaranteed here: https://github.com/crt25/collimator/pull/576/changes#diff-2295d3d0d6c3cc9322d3e39c566fa8b56d77fb683695db1654bd96fd2881b86eR40

pierluca (approved):

PR #582, opened 2026-06-01
[CRT-418] HTTP 500 - timeout when selecting reference - missing index

bubudetp:
## Summary by cubic
Fixes CRT-418 by indexing `StudentActivity` queries to stop HTTP 500 timeouts when selecting a reference. Solution-hash lookups are now fast and do not block transactions.

- **Bug Fixes**
  - Add Prisma `@@index([studentId, solutionHash])` and `@@index([taskId, solutionHash])` on `StudentActivity` to speed `updateStudentActivityIsReference` and joins in `getCurrentAnalysesWithActivities`.

- **Migration**
  - Apply the Prisma migration to create both indexes.

<sup>Written for commit e44542b90f19928693101c49bb57c5d4344a8c64. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/crt25/collimator/pull/582?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/prisma/schema.prisma
+++ b/backend/prisma/schema.prisma
@@ -353,7 +353,8 @@ model StudentActivity {
   appActivity StudentActivityApp?
 
   @@unique([studentId, type, happenedAt, happenedAtCounter], name: "uniqueStudentActivityPerTypeAndTime")
-  @@index([solutionHash])
+  @@index([studentId, solutionHash])
+  @@index([taskId, solutionHash])
 }
 
 model StudentActivityApp {
--- a/backend/prisma/schema.prisma
+++ b/backend/prisma/schema.prisma
@@ -353,6 +353,7 @@ model StudentActivity {
   appActivity StudentActivityApp?
 
   @@unique([studentId, type, happenedAt, happenedAtCounter], name: "uniqueStudentActivityPerTypeAndTime")
+  @@index([solutionHash])
 }
 
 model StudentActivityApp {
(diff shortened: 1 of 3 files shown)


backend/prisma/schema.prisma
```diff
@@ -353,6 +353,7 @@ model StudentActivity {
   appActivity StudentActivityApp?
 
   @@unique([studentId, type, happenedAt, happenedAtCounter], name: "uniqueStudentActivityPerTypeAndTime")
+  @@index([solutionHash])
```
pierluca: do we ever access `solutionHash` without providing any other fields? 

If not I'd suggest indexing a combination of fields, e.g. studentId, solutionHash, and perhaps happenedAt+happenedAtCounter so that we can retrieve the latest 🤔

bubudetp: A studentId + solutionHash is the better fix for the failing transaction. Also adding taskId + solutionHash seems more appropriate for the analyses with activities join there's no need for the happenedAt+happenedAtCounter because the query has no order by so it should be enough.

pierluca: Ok, still awaiting the fix :-)

Re: happenedAt / happenedAtCounter , do we really not need to distinguish between the various identical solutions? when are they fungible? (for my own understanding)

bubudetp: Hmm there's no read query on happenedAt+happenedAtCounter anywhere in the codebase, they are only used for the activity deduplication unique constraint which already has its own index ([see here](https://github.com/crt25/collimator/blob/main/backend/prisma/schema.prisma#L355)). In getCurrentAnalysesWithActivities.sql we are returning the latest activity per student by ordering by createdAt which is the right choice anyway because happenedAt comes from the student activity.

pierluca (approved):

PR #583, opened 2026-06-01
[CRT-414] Scratch doesn't reset to correct position after reloading the task

bubudetp:
## Summary by cubic
Fixes sprites and stage not resetting after reload by snapshotting the start state on Green Flag and using it during serialization. Addresses CRT-414 so runtime changes no longer redefine the saved starting point.

- Bug Fixes
  - Snapshot stage and original sprites on Green Flag before scripts run; exclude clones; clear snapshot on project load.
  - On save, temporarily restore start state (costume, position, size, direction, visible, layer order, volume, tempo, draggable, rotation style, custom state), then restore live runtime.
  - Add `vm/target-state` helpers and use them in assertions to reset state for test execution.
  - Deep-copy `_customState` during snapshot/restore to avoid mutation bleed.
  - Align with `@scratch/scratch-vm`: `loadExtensionURL` returns `0`; export `Target` and `PostedSpriteInfo` types; default `enableStageInteractions` to false in config and UI.

<sup>Written for commit b726f6fdb22e66128a4a8277d4dfe6ccde4ca3bb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/crt25/collimator/pull/583?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/scratch/src/vm/index.ts
+++ b/apps/scratch/src/vm/index.ts
@@ -3,7 +3,6 @@ import { ExtensionId } from "../extensions";
 import AssertionExtension from "../extensions/assertions";
 import {
   RememberedSpriteState,
-  refreshSpriteEditorState,
   rememberSpriteState,
   restoreSpriteStateForSerialization,
 } from "./target-state";
@@ -148,32 +147,8 @@ const wrapPostSpriteInfo = (vm: VM): void => {
   };
 };
 
-const mirrorEditsOnTargetsUpdate = (vm: VM): void => {
-  const startState = getTargetStartState(vm);
-  // rename and costume changes trigger a TARGETS_UPDATE
-  vm.runtime.on("TARGETS_UPDATE", () => {
-    if (vm.runtime.threads.length > 0) {
-      return;
-    }
-
-    for (const target of vm.runtime.targets) {
-      // skip original and stage targets
-      if (!isTrackableSprite(target)) {
-        continue;
-      }
-
-      const snap = startState.get(target.id);
-
-      if (snap) {
-        refreshSpriteEditorState(snap, target);
-      }
-    }
-  });
-};
-
 const wrapEditorMutators = (vm: VM): void => {
   wrapPostSpriteInfo(vm);
-  mirrorEditsOnTargetsUpdate(vm);
 };
 
 const snapshotOnTargetLifecycle = (vm: VM): void => {
--- a/apps/scratch/src/vm/index.ts
+++ b/apps/scratch/src/vm/index.ts
@@ -3,54 +3,80 @@ import { ExtensionId } from "../extensions";
 import AssertionExtension from "../extensions/assertions";
 import {
   RememberedSpriteState,
+  RememberedStageState,
   rememberSpriteState,
+  rememberStageState,
   restoreSpriteStateForExecution,
   restoreSpriteStateForSerialization,
+  restoreStageStateForExecution,
+  restoreStageStateForSerialization,
 } from "./target-state";
 
-const startStateByVm = new Map<VM, Map<string, RememberedSpriteState>>();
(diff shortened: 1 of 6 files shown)


apps/scratch/src/vm/index.ts
```diff
@@ -40,5 +139,53 @@ export const patchScratchVm = (vm: VM): void => {
...
+
+    if (!target || target.isStage) {
+      return;
+    }
+
+    const snap = startState.get(target.id);
+
+    if (!snap) {
+      return;
+    }
```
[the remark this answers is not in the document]
bubudetp: Good catch but this solution wouldn't prevent against the flow where the user clicks on "choose a sprite" and the sprite is set at the default position. The user then writes code, executes it and saves because it won't hit postSpriteInfo so the serializer will still write the last pos of the executed code when saving

pierluca (changes requested):
I found where we already had most of the (similar) logic:
apps/scratch/src/extensions/assertions/index.ts

That might be what we somehow need to fix and/or integrate with the saving 😅

apps/scratch/src/vm/index.ts
```diff
@@ -1,6 +1,75 @@
...
+    if (!start) {
+      continue;
+    }
+
+    runtimeState.push({
+      target,
+      runtime: rememberSpriteState(target),
+    });
+
+    restoreSpriteStateForSerialization(target, start);
```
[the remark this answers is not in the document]
bubudetp: Addressed in #586

apps/scratch/src/vm/index.ts
```diff
@@ -1,6 +1,71 @@
...
+const startStateByVm = new Map<VM, Map<string, RememberedSpriteState>>();
+
+/**
+ * Snapshot on project load, on `targetWasCreated`, and on user-initiated
+ * postSpriteInfo calls (stage drag, sprite-info pane). Used by the
+ * vm.toJSON wrap below so that running a script and then serialising
+ * doesn't introduce runtime differences into the
+ * project as the new starting point.
+ */
+export const getTargetStartState = (
```
pierluca:
```suggestion
export const getSpritesStartState = (
```

seems like this only refers to sprites?

apps/scratch/src/vm/index.ts
pierluca, line 1: Not saving the stage here?

apps/scratch/src/vm/index.ts
```diff
@@ -40,5 +115,77 @@ export const patchScratchVm = (vm: VM): void => {
...
+      snap.visible = target.visible;
+    }
+
+    if (typeof data.draggable === "boolean") {
+      snap.draggable = target.draggable;
+    }
+
+    if (typeof data.rotationStyle === "string") {
+      snap.rotationStyle = target.rotationStyle;
+    }
```
pierluca: why do we need these checks? 🤔

apps/scratch/src/vm/index.ts
bubudetp: As I said, saving the stage is in a subsequent PR: #586

apps/scratch/src/vm/index.ts
bubudetp: Both are correct but your suggestion is more precise 😅

apps/scratch/src/vm/index.ts
bubudetp: It's because these fields are optional

pierluca (approved):

PR #584, opened 2026-06-01
[CRT-413] Student don't show up in progress report until they click "Submit"

bubudetp:
## Summary by cubic
Fixes progress reports so students who start an analysis appear and are marked In Progress before submitting, and auto-refreshes the list after submit. Addresses CRT-413.

- **Bug Fixes**
  - Add `useAllSessionCurrentAnalyses` in progress lists to include students with active analyses across tasks.
  - Update status/color and deps: analyses-only -> In Progress (error); memo deps include `currentAnalyses`/`rowData.currentAnalysis`; `findAnalysisToDisplay` prefers latest non-starred (fallback to starred).
  - Revalidate cached “current analyses” after submit via `useRevalidateSolutionList`; reuse `allTasksPlaceholder` when building current-analyses URLs.

<sup>Written for commit 65d8ae08e614ab3d0ca91be3cc1a5fc03e8fc275. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/crt25/collimator/pull/584?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/src/components/task-instance/TaskInstanceProgressList.tsx
+++ b/frontend/src/components/task-instance/TaskInstanceProgressList.tsx
@@ -115,11 +115,13 @@ const TaskTemplate = ({
   );
 
   const status = useMemo(() => {
-    if (!solutionToDisplay) {
+    if (!solutionToDisplay && !rowData.currentAnalysis) {
       // if solution has an analysis then it should be displayed as in progress
-      return rowData.currentAnalysis
-        ? TaskStatus.incomplete
-        : TaskStatus.notStarted;
+      return TaskStatus.notStarted;
+    }
+
+    if (!solutionToDisplay) {
+      return TaskStatus.incomplete;
     }
 
     if (solutionToDisplay.tests.every((test) => test.passed)) {
@@ -130,8 +132,13 @@ const TaskTemplate = ({
   }, [solutionToDisplay, rowData.currentAnalysis]);
 
   const color = useMemo((): StatusColor => {
+    if (!solutionToDisplay && !rowData.currentAnalysis) {
+      // if solution has an analysis then it should be displayed as in progress
+      return "neutral";
+    }
+
     if (!solutionToDisplay) {
-      return rowData.currentAnalysis ? "error" : "neutral";
+      return "error";
     }
 
     if (solutionToDisplay.tests.every((test) => test.passed)) {
--- a/frontend/src/components/task-instance/TaskInstanceProgressList.tsx
+++ b/frontend/src/components/task-instance/TaskInstanceProgressList.tsx
@@ -240,9 +240,9 @@ const TaskInstanceProgressList = ({
       );
 
       // A student is either already in the showcase (teacher needs to unstar)
-      // or not yet (teacher needs to star). findShowcaseAnalysis prefers the
-      // starred analysis so the button reflects the current state and allows
-      // removal; falls back to the latest non-reference analysis to add it.
+      // or not yet (teacher needs to star). findAnalysisToDisplay returns the
+      // latest non-starred analysis so the button is starrable.
+      // It falls back to the starred analysis when that's all the student has.
       const currentAnalysis = CurrentStudentAnalysis.findAnalysisToDisplay(
         currentAnalyses ?? [],
         student.studentId,
--- a/frontend/src/components/task-instance/TaskInstanceProgressList.tsx
+++ b/frontend/src/components/task-instance/TaskInstanceProgressList.tsx
@@ -116,7 +116,10 @@ const TaskTemplate = ({
 
   const status = useMemo(() => {
     if (!solutionToDisplay) {
-      return TaskStatus.notStarted;
+      // if solution has an analysis then it should be displayed as in progress
+      return rowData.currentAnalysis
+        ? TaskStatus.incomplete
+        : TaskStatus.notStarted;
     }
 
     if (solutionToDisplay.tests.every((test) => test.passed)) {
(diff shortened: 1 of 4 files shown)


frontend/src/components/task-instance/TaskInstanceProgressList.tsx
pierluca: out of scope for the current PR, but isn't this comment out of date? 🤔 
It looks like we're not using `findShowcaseAnalysis`

frontend/src/components/task-instance/TaskInstanceProgressList.tsx
```diff
@@ -204,9 +204,15 @@ const TaskInstanceProgressList = ({
...
+      ...(currentAnalyses ?? [])
+        .filter(
+          (a): a is CurrentStudentAnalysis =>
+            a instanceof CurrentStudentAnalysis,
+        )
+        .map((a) => a.studentId),
     ]);
     return [...studentIdsSet];
-  }, [klass, solutions]);
+  }, [klass, solutions, currentAnalyses]);
```
pierluca, line 217: this will add their `id` to the list, but how do we make the distinction between `Not started` and `in progress` ? I don't see any changes in the column rendering 🤔 what am I missing?

bubudetp: No you're right, I forgot to adapt the rows as well as the colors. We can make the distinction between not started and in progress by checking if the solution has an analysis.

frontend/src/components/task-instance/TaskInstanceProgressList.tsx
```diff
@@ -116,15 +116,18 @@ const TaskTemplate = ({
 
   const status = useMemo(() => {
     if (!solutionToDisplay) {
-      return TaskStatus.notStarted;
+      // if solution has an analysis then it should be displayed as in progress
+      return rowData.currentAnalysis
```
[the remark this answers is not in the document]
pierluca: I agree with the bot.

Also, it might make sense for coherency with the rest of the function design to have something like 

```ts
    if (!solutionToDisplay && !rowData.currentAnalysis) {
      // if solution has an analysis then it should be displayed as in progress
      return TaskStatus.notStarted
    }
```

frontend/src/components/task-instance/TaskInstanceProgressList.tsx
bubudetp: good catch, and yeah I must have forgotten to update it after a refactor

frontend/src/components/task-instance/TaskInstanceProgressList.tsx
```diff
@@ -231,9 +240,9 @@ const TaskInstanceProgressList = ({
       );
 
       // A student is either already in the showcase (teacher needs to unstar)
-      // or not yet (teacher needs to star). findShowcaseAnalysis prefers the
-      // starred analysis so the button reflects the current state and allows
-      // removal; falls back to the latest non-reference analysis to add it.
```
pierluca, line 236: was this change in behavior desired in a ticket though? Or is this a side effect of logic changing? I can't help but wonder 🤔 

As a teacher, if the student is editing the task, will I be tempted to star it many many times just because the button doesn't turn yellow?

bubudetp: The old logic always showed the most recent solution regardless of the showcase status, the new one prioritizes showing the unstarred one, which means that if a student has a starred solution and keeps submitting new ones, the teacher sees the latest unstarred work and star the button which would show as gray and I have to agree with you that it can be confusing. 

But what is the intended ux then? if our goal is to show the teacher the latest work they can decide whether to star it then the new behavior is the correct one, but the star button should probably reflect whether any solution is already starred to avoid the confusion. 

I can suggest to have a filled star in yellow to represent if any of the student's solutions isReferneceSolution === true.

pierluca: As per our discussion earlier this week, do briefly summarize this and submit a question to Engin. Feel free to ask @AthenaPK for a second pair of eyes before sending, as this can be quite hard to grasp with a cold cache :-)

frontend/src/components/task-instance/TaskInstanceProgressList.tsx
```diff
@@ -115,28 +115,38 @@ const TaskTemplate = ({
...
+    if (!solutionToDisplay && !rowData.currentAnalysis) {
+      // if solution has an analysis then it should be displayed as in progress
       return "neutral";
     }
 
+    if (!solutionToDisplay) {
+      return "error";
+    }
+
     if (solutionToDisplay.tests.every((test) => test.passed)) {
```
pierluca:
```suggestion
    if (solutionToDisplay && solutionToDisplay.tests.every((t) => t.passed)) {
```

frontend/src/components/task-instance/TaskInstanceProgressList.tsx
```diff
@@ -115,28 +115,38 @@ const TaskTemplate = ({
...
+    if (!solutionToDisplay && !rowData.currentAnalysis) {
+      // if solution has an analysis then it should be displayed as in progress
       return TaskStatus.notStarted;
     }
 
+    if (!solutionToDisplay) {
+      return TaskStatus.incomplete;
+    }
+
     if (solutionToDisplay.tests.every((test) => test.passed)) {
```
pierluca:
```suggestion
    if (solutionToDisplay && solutionToDisplay.tests.every((t) => t.passed)) {
```

pierluca (approved):

pierluca (approved):

PR #586, opened 2026-06-09
[CRT-414] Preserve stage & sprite metadata on update

bubudetp:
## Summary by cubic
Fixes CRT-414 by snapshotting the stage and original sprites on the first green flag (before hat threads) and using that only during serialization so projects reload with the intended setup. Also defaults `enableStageInteractions` to false to avoid unintended state changes.

- **Bug Fixes**
  - Snapshot stage + original sprites once on the first green flag; reuse on subsequent runs and clear on `PROJECT_LOADED`.
  - During `vm.toJSON`, temporarily apply start state via `resetToStartState`, then restore runtime (restores both stage and sprites).
  - Track only original sprites (skip clones) and capture full metadata (names, costume, volume, x/y, size, direction, visibility, draggable, rotation style, layer, tempo, video transparency, custom state). Skip targets without saved state.
  - Fix inverted clone filter in `backupTargetsState` so clones are skipped correctly.

- **Refactors**
  - Grouped helpers and extracted `backupTargetsState`; `patchSerialization` now uses `resetToStartState` (renamed from `swapToStartState`). Scoped start state inside `patchScratchVm` and removed `targetWasCreated`/`postSpriteInfo` listeners; snapshot is driven by the run lifecycle.
  - Inlined the “trackable sprite” condition in `backupTargetsState`.

<sup>Written for commit a8e0face6b40bd274be9d35c344f0d351991119c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/crt25/collimator/pull/586?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/scratch/src/vm/index.ts
+++ b/apps/scratch/src/vm/index.ts
@@ -3,7 +3,6 @@ import { ExtensionId } from "../extensions";
 import AssertionExtension from "../extensions/assertions";
 import {
   RememberedSpriteState,
-  refreshSpriteEditorState,
   rememberSpriteState,
   restoreSpriteStateForSerialization,
 } from "./target-state";
@@ -148,32 +147,8 @@ const wrapPostSpriteInfo = (vm: VM): void => {
   };
 };
 
-const mirrorEditsOnTargetsUpdate = (vm: VM): void => {
-  const startState = getTargetStartState(vm);
-  // rename and costume changes trigger a TARGETS_UPDATE
-  vm.runtime.on("TARGETS_UPDATE", () => {
-    if (vm.runtime.threads.length > 0) {
-      return;
-    }
-
-    for (const target of vm.runtime.targets) {
-      // skip original and stage targets
-      if (!isTrackableSprite(target)) {
-        continue;
-      }
-
-      const snap = startState.get(target.id);
-
-      if (snap) {
-        refreshSpriteEditorState(snap, target);
-      }
-    }
-  });
-};
-
 const wrapEditorMutators = (vm: VM): void => {
   wrapPostSpriteInfo(vm);
-  mirrorEditsOnTargetsUpdate(vm);
 };
 
 const snapshotOnTargetLifecycle = (vm: VM): void => {
--- a/apps/scratch/src/vm/index.ts
+++ b/apps/scratch/src/vm/index.ts
@@ -3,54 +3,80 @@ import { ExtensionId } from "../extensions";
 import AssertionExtension from "../extensions/assertions";
 import {
   RememberedSpriteState,
+  RememberedStageState,
   rememberSpriteState,
+  rememberStageState,
   restoreSpriteStateForExecution,
   restoreSpriteStateForSerialization,
+  restoreStageStateForExecution,
+  restoreStageStateForSerialization,
 } from "./target-state";
 
-const startStateByVm = new Map<VM, Map<string, RememberedSpriteState>>();
+const spriteStartStateByVm = new Map<VM, Map<string, RememberedSpriteState>>();
+const stageStartStateByVm = new Map<VM, RememberedStageState>();
 
-export const getTargetStartState = (
+export const getSpriteStartState = (
   vm: VM,
 ): Map<string, RememberedSpriteState> => {
-  let map = startStateByVm.get(vm);
+  let map = spriteStartStateByVm.get(vm);
 
   if (!map) {
     map = new Map();
-    startStateByVm.set(vm, map);
+    spriteStartStateByVm.set(vm, map);
   }
 
   return map;
 };
 
+export const getStageStartState = (vm: VM): RememberedStageState | undefined =>
+  stageStartStateByVm.get(vm);
+
 const clearStartState = (vm: VM): void => {
-  getTargetStartState(vm).clear();
+  getSpriteStartState(vm).clear();
+  stageStartStateByVm.delete(vm);
 };
 
-/**
- * Temporarily set each target's mutable sprite state to its tracked
- * starting state so the serializer doesn't capture runtime drift caused
- * by scripts. Returns a restore callback.
- */
+const isStageWithStartState = (
+  target: VM.Target,
+  vm: VM,
+  stage: RememberedStageState | undefined,
+): stage is RememberedStageState =>
+  target.isStage && stageStartStateByVm.has(vm) && stage !== undefined;
+
 export const swapToStartState = (vm: VM): (() => void) => {
-  const snapshot = getTargetStartState(vm);
-  if (snapshot.size === 0) {
+  const sprites = getSpriteStartState(vm);
+  const stage = getStageStartState(vm);
+
+  // if there is no stage or trackable sprites with start state, we don't need to do anything
+  if (sprites.size === 0 && !stage) {
     return () => {};
   }
 
-  const runtimeState: Array<{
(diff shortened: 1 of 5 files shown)


apps/scratch/src/vm/index.ts
```diff
@@ -3,56 +3,78 @@ import { ExtensionId } from "../extensions";
...
 
-/**
- * Snapshot on project load, on `targetWasCreated`, and on user-initiated
- * postSpriteInfo calls (stage drag, sprite-info pane). Used by the
- * vm.toJSON wrap below so that running a script and then serialising
- * doesn't introduce runtime differences into the
- * project as the new starting point.
- */
-export const getTargetStartState = (
+export const getSpriteStartState = (
```
pierluca: similar comment as the other PR 😅 this is collecting _all_ sprites right? 

`spriteStartStateByVm` -> `spritesStartStateBym`
`getSpriteStartState` -> `getSpritesStartState`

apps/scratch/src/vm/index.ts
```diff
@@ -3,56 +3,78 @@ import { ExtensionId } from "../extensions";
...
+  RememberedStageState,
   rememberSpriteState,
+  rememberStageState,
   restoreSpriteStateForSerialization,
+  restoreStageStateForSerialization,
 } from "./target-state";
 
-const startStateByVm = new Map<VM, Map<string, RememberedSpriteState>>();
+const spriteStartStateByVm = new Map<VM, Map<string, RememberedSpriteState>>();
+const stageStartStateByVm = new Map<VM, RememberedStageState>();
```
pierluca: for my understanding, why does this need to be indexed on the VM?
When do we expect to have multiple competing VMs? 🤔

apps/scratch/src/vm/index.ts
```diff
@@ -3,56 +3,59 @@ import { ExtensionId } from "../extensions";
...
+      stageRuntime = { target, runtime: rememberStageState(target) };
+      restoreStageStateForSerialization(target, startState.stage);
+      continue;
+    }
+
+    const start = startState.sprites.get(target.id);
 
     if (!start) {
       continue;
     }
```
pierluca: wait, if the sprite didn't exist at the starting point, shouldn't we exclude it from serialization? 🤔

apps/scratch/src/vm/index.ts
```diff
@@ -116,71 +124,40 @@ export const patchScratchVm = (vm: VM): void => {
...
-    }
+        if (!isTrackableSprite(target)) {
+          continue;
+        }
 
-    if (typeof data.draggable === "boolean") {
-      snap.draggable = target.draggable;
+        startState.sprites.set(target.id, rememberSpriteState(target));
+      }
     }
```
pierluca: This whole section could be its own `backupTargetsState()` helper function.

apps/scratch/src/vm/index.ts
```diff
@@ -116,71 +124,40 @@ export const patchScratchVm = (vm: VM): void => {
...
-      snap.visible = target.visible;
-    }
+        if (!isTrackableSprite(target)) {
+          continue;
+        }
 
-    if (typeof data.draggable === "boolean") {
-      snap.draggable = target.draggable;
+        startState.sprites.set(target.id, rememberSpriteState(target));
+      }
```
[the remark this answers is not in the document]
pierluca: I agree, why do we need line 135 here? 

In principle we could just call `clearStartState(startState);` at the beginning of the green flag 🤔

apps/scratch/src/vm/index.ts
```diff
@@ -116,71 +124,40 @@ export const patchScratchVm = (vm: VM): void => {
...
-  // callers serialise runtime drift caused by scripts and end up sending
-  // the runtime sprite position to the backend as the new starting point.
+// temporarily swap to start state during serialization to ensure
+// the project is saved with original positions, not changes occured in runtime.
+// immediately restore the current runtime state after serialization.
+const patchSerialization = (vm: VM, startState: StartState): void => {
   const originalToJSON = vm.toJSON.bind(vm);
   vm.toJSON = (optTargetId?: string): string => {
-    const restore = swapToStartState(vm);
+    const restore = swapToStartState(vm, startState);
```
pierluca:
```suggestion
    const undo = revertToStartState(vm, startState);
```
or 
```suggestion
    const undo = resetToStartState(vm, startState);
```

would be super clear

pierluca (commented):
Sorry for these late comments :-(

apps/scratch/src/vm/index.ts
```diff
@@ -3,56 +3,59 @@ import { ExtensionId } from "../extensions";
...
+  const spriteRuntimeState: Array<{
     target: VM.Target;
     runtime: RememberedSpriteState;
   }> = [];
 
+  let stageRuntime: {
+    target: VM.Target;
+    runtime: RememberedStageState;
+  } | null = null;
+
```
pierluca: Maybe a good idea, maybe not:
if we have a `backupTargetsState()` as defined in the other comment, would it make sense to reuse it here and just have a separate, complementary `resetTargetsForSerialization()` function ? 

We'd be iterating twice, but maybe sharing more logic with fewer chances of bugs? 🤔

apps/scratch/src/vm/index.ts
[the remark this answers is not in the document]
bubudetp: We need it in the scenario where the script just ran and sprite went from (0,0) to (100,0) the sprite does not get reset to its initial position. The student can press the greenflag button, and without this guard, (100,0) will be snapshotted as the start state and on new execution the sprite's initial position will be reset to (100, 0) and then reach (200,0) in the next iteration. Does this make sense?

apps/scratch/src/vm/index.ts
```diff
@@ -2,72 +2,55 @@ import VM from "@scratch/scratch-vm";
...
       continue;
     }
 
-    runtimeState.push({
-      target,
-      runtime: rememberSpriteState(target),
-    });
+    const sprite = state.sprites.get(target.id);
+    if (!sprite) {
+      return;
```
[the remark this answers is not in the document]
bubudetp: That's the point, why serialize sprites if no sprites are in the runtime?

apps/scratch/src/vm/target-state.ts
```diff
@@ -117,3 +117,28 @@ export const restoreSpriteStateForSerialization = (
...
+  }
+
+  for (const target of vm.runtime.targets) {
+    if (target.isStage) {
+      startState.stage = rememberStageState(target);
+      continue;
+    }
+
+    // track every sprite the project file defines whether shown or hidden, skip the stage and runtime clones.
+    if (!target.isStage && target.isOriginal !== false) {
```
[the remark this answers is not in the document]
bubudetp: Good catch

apps/scratch/src/vm/index.ts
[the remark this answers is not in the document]
bubudetp: NVM, you were right

pierluca (approved):

PR #592, opened 2026-06-11
[CRT-425] Cache and reuse task parent origin

bubudetp:
## Summary by cubic
Cache and reuse the parent origin in `useIframeParent` so RPC messages aren’t rejected after changing a scratch task’s language (CRT-425).

- **Bug Fixes**
  - Cache the parent origin at module scope, initialize `taskOrigin` from it, and document the reload-gap workaround.
  - Wrap `loadSubmission` and `loadTask` to refresh both the cache and state when a new origin is provided.

<sup>Written for commit 74e3f31b64b3c0abe37896407cef521da541fe69. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/crt25/collimator/pull/592?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/libraries/iframe-rpc-react/src/useIframeParent.ts
+++ b/libraries/iframe-rpc-react/src/useIframeParent.ts
@@ -8,6 +8,8 @@ import {
 import { ParametersOf } from "iframe-rpc/src/utils";
 import { useEffect, useMemo, useRef, useState } from "react";
 
+let cachedParentOrigin: string | null = null;
+
 export const useIframeParent = (
   handleRequest: AppHandleRequestMap | null,
   initialMessages?: MessageEvent<unknown>[],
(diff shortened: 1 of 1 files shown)


libraries/iframe-rpc-react/src/useIframeParent.ts
```diff
@@ -8,6 +8,8 @@ import {
 import { ParametersOf } from "iframe-rpc/src/utils";
 import { useEffect, useMemo, useRef, useState } from "react";
 
+let cachedParentOrigin: string | null = null;
```
pierluca, line 15:
```suggestion
// We cache the parent origin as a workaround to prevent transient, reload bugs.
// During an iframe reload, no RPC is called by the parent, which in turn means
// that no origin is set in the crtPlatform and we become unable to communicate
// with the parent. This fixes CRT-425 and similar classes of bugs.
let cachedParentOrigin: string | null = null;
```

pierluca (approved):

PR #594, opened 2026-06-11
[CRT-411] Snapshot task block ids & reset after import export roundtrip

bubudetp:
## Summary by cubic
Prevents blocks from becoming undeletable after an export/import roundtrip by snapshotting and restoring each block’s `isTaskBlock` (including `undefined`) around the reload, even if it fails. Addresses CRT-411 so task marks stay correct and blocks remain editable.

- **Bug Fixes**
  - Snapshot `isTaskBlock` for all runtime blocks before `loadCrtProject` and restore in a `try/finally` after the reload or failure.
  - Neutralizes `PROJECT_LOADED` side effects by restoring pre-reload values so workspace blocks aren’t locked as task blocks.

- **Refactors**
  - Added `iterateAllBlocks` and derived the `BlockExtended` type from the VM for safer traversal; used in load init and the export roundtrip.

<sup>Written for commit b5199fa5f35a61c7a6f425b7bd0ca23eacfba5bf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/crt25/collimator/pull/594?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/scratch/src/vm/utils.ts
+++ b/apps/scratch/src/vm/utils.ts
@@ -1,8 +1,6 @@
-import { BlockExtended } from "@scratch/scratch-vm";
-
 export function* iterateAllBlocks(
   runtime: VMExtended.RuntimeExtended,
-): Generator<BlockExtended> {
+): Generator<VMExtended.BlockExtended> {
   for (const target of runtime.targets) {
     for (const block of Object.values(target.blocks._blocks)) {
       yield block;
--- a/apps/scratch/src/vm/utils.ts
+++ b/apps/scratch/src/vm/utils.ts
@@ -1,0 +1,11 @@
+import { BlockExtended } from "@scratch/scratch-vm";
+
+export function* allBlocks(
+  runtime: VMExtended.RuntimeExtended,
+): Generator<BlockExtended> {
+  for (const target of runtime.targets) {
+    for (const block of Object.values(target.blocks._blocks)) {
+      yield block;
+    }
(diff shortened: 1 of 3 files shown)


pierluca (commented):
The CI fails, but the code doesn't look bad to me.

Perhaps out of scope for this PR, but it seems that we now have a bunch of places where we iterate over all targets and blocks. Would it make sense to have a small utility function such that we can do something like
`iterateBlocks(vm.runtime).forEach((target, block) => { ... })` and re-use in all those places?

apps/scratch/src/vm/utils.ts
```diff
@@ -0,0 +1,11 @@
+import { BlockExtended } from "@scratch/scratch-vm";
+
+export function* allBlocks(
```
pierluca: `all blocks` is a noun phrase, not a verb.
As such, it doesn't really indicate what the action is.
I would strive to have functions with a verb in them, e.g. `iterateAllBlocks` or something similar.

pierluca (approved):
Minor suggestion, but LGTM

PR #595, opened 2026-06-12
[CRT-412] Only consider non reference entries in progress list

bubudetp:
## Summary by cubic
Fixes duplicate students in the progress report by excluding reference solutions when building the list (CRT-412). Replaces `unique-names-generator` with a deterministic nickname generator based on `studentId` and `locale`, decorrelates animal/number indices, and removes use of pseudonyms entirely.

<sup>Written for commit 377a1d47528827f3f80d620a21be74f371161d85. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/crt25/collimator/pull/595?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/src/utilities/student-name.ts
+++ b/frontend/src/utilities/student-name.ts
@@ -1,4 +1,3 @@
-import { uniqueNamesGenerator } from "unique-names-generator";
 import { Language } from "iframe-rpc-react/src";
 import animals from "../../content/dictionaries/animals";
 import numbers from "../../content/dictionaries/numbers";
@@ -11,14 +10,12 @@ export function getStudentNickname(
   // We would prefer prefixing the pseudonym to avoid the two name spaces
   // from clashing.
   // However for backwards compatibility we don't, see https://github.com/crt25/collimator/pull/286/files#r2054086764.
-  const seed = pseudonym ? pseudonym : `anonymous_${studentId}`;
-  const animalNames = animals.map((animal) => animal[locale]);
-
-  return uniqueNamesGenerator({
-    dictionaries: [animalNames, numbers],
-    separator: "-",
-    length: 2,
-    style: "lowerCase",
-    seed,
-  });
+  return pseudonym ? pseudonym : generateAnonymousName(studentId, locale);
 }
+
+const generateAnonymousName = (studentId: number, locale: Language): string => {
+  const animalIndex = studentId % animals.length;
+  const numberIndex = studentId % numbers.length;
+
+  return `${animals[animalIndex][locale]}-${numbers[numberIndex]}`;
+};
--- a/frontend/src/utilities/student-name.ts
+++ b/frontend/src/utilities/student-name.ts
@@ -4,15 +4,11 @@ import numbers from "../../content/dictionaries/numbers";
 
 export function getStudentNickname(
   studentId: number,
-  pseudonym?: string | null,
   locale: Language = Language.en,
 ): string {
-  // We would prefer prefixing the pseudonym to avoid the two name spaces
-  // from clashing.
-  // However for backwards compatibility we don't, see https://github.com/crt25/collimator/pull/286/files#r2054086764.
-  const seed = pseudonym ? pseudonym : `anonymous_${studentId}`;
-  const hash = hashString(seed);
-  const animalIndex = hash % animals.length;
+  const animalIndex = studentId % animals.length;
+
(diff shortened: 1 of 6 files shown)


frontend/src/components/dashboard/ProgressList.tsx
pierluca, line 221: Seems unlikely to be the issue. `studentId` goes through a `new Set()`, so duplicates would've been removed.

pierluca, line 248: could multiple solutions be present here?

pierluca (changes requested):
Not sure what the bug is, but I assume

frontend/src/utilities/student-name.ts
```diff
@@ -11,14 +10,12 @@ export function getStudentNickname(
...
-    length: 2,
-    style: "lowerCase",
-    seed,
-  });
+  return pseudonym ? pseudonym : generateAnonymousName(studentId, locale);
 }
+
+const generateAnonymousName = (studentId: number, locale: Language): string => {
+  const animalIndex = studentId % animals.length;
+  const numberIndex = studentId % numbers.length;
```
pierluca: The number doesn't serve its purpose here, which is to extend the range of possible pseudonyms.

we'll have

{animal1}-1
{animal2}-2
{animal3}-3
{animal4}-4
{animal5}-5
etc.

frontend/src/utilities/student-name.ts
```diff
@@ -11,14 +10,12 @@ export function getStudentNickname(
...
-  const animalNames = animals.map((animal) => animal[locale]);
-
-  return uniqueNamesGenerator({
-    dictionaries: [animalNames, numbers],
-    separator: "-",
-    length: 2,
-    style: "lowerCase",
-    seed,
-  });
+  return pseudonym ? pseudonym : generateAnonymousName(studentId, locale);
```
pierluca: This changes completely the behaviour. Was it incorrect?

From the look of the function though, I'm wondering if we shouldn't always use `studentId` instead of the pseudonym.

pierluca (changes requested):
Haven't had time to look deep into the specs, but this has some clear issues.

frontend/src/utilities/student-name.ts
bubudetp: We derive the pseudonym from the student's identity, I think we should keep it

frontend/src/utilities/student-name.ts
```diff
@@ -12,13 +11,18 @@ export function getStudentNickname(
...
+  const animalIndex = hash % animals.length;
+  const numberIndex = hash % numbers.length;
 
-  return uniqueNamesGenerator({
-    dictionaries: [animalNames, numbers],
-    separator: "-",
-    length: 2,
-    style: "lowerCase",
-    seed,
-  });
```
pierluca, line 23: Looked into the problem here, basically the library has a weakness and we were using it in a problematic way.

The `seed` was supposed to be a number, but a string input is tolerated.

Inside, the library does `.split('').reduce((acc, curr) => acc + curr.charCodeAt(0), 1)` on the string to convert it into a numeric seed [[see here]](https://github.com/andreasonny83/unique-names-generator/blob/main/src/unique-names-generator.constructor.ts#L39-L41), which means the individual characters get summed up. `prefix_329` and `prefix_338` sum up to the same number, because 2+9 and 3+8 are the same.

From then on, the things become identical.

Solution: pass a number instead of a string. We already have `studentId`, job done ;)

Now, since (as you pointed out correctly), this is a dependency that we may not really need, we could also keep your current logic, and decorrelate further animalIndex/numberIndex by doing something like:

```
  const animalIndex = studentId % animals.length;

  const hash = hashString(studentId);
  const numberIndex = hash % numbers.length;

  return `${animals[animalIndex][locale]}-${numbers[numberIndex]}`;
```

This way students that were added close to one another (sequential IDs) are pretty much guaranteed not to use the same animal, yet the number doesn't systematically grow at the same time.

pierluca (approved):

PR #596, opened 2026-06-25
[CRT-401] Reference solutions do not allow clicking "save" unless descriptions are filled

bubudetp:
image

---
## Summary by cubic
Aligns Reference Solutions with the Task form: Save is enabled when the form is edited, and validation runs on submit with clear inline errors. Addresses CRT-401.

- **Bug Fixes**
  - Save button is enabled when the form is dirty; validation runs on submit only.
  - Inline errors show under each solution’s title and description.
  - After first submit, field edits validate only that field; removing a solution or changing a file re-validates the full form to clear stale errors.
  - Fixed submission to invoke the handler returned by `handleSubmit(e)`, so Save runs validation and prevents native browser submission.
  - Clarified the file-missing message and updated `en`/`fr` strings; improved save error handling with try/catch and toasts for duplicate solutions and generic failures.

<sup>Written for commit 8c9311add76378ed9857c3134218e0b34a647174. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/crt25/collimator/pull/596?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/src/components/task/TaskFormReferenceSolutions.tsx
+++ b/frontend/src/components/task/TaskFormReferenceSolutions.tsx
@@ -286,9 +286,8 @@ const TaskFormReferenceSolutions = ({
       setValue("referenceSolutions", referenceSolutions, {
         shouldDirty: true,
       });
-      revalidate();
     },
-    [setValue, revalidate],
+    [setValue],
   );
 
   const updateReferenceSolution = useCallback(
--- a/frontend/src/components/task/TaskFormReferenceSolutions.tsx
+++ b/frontend/src/components/task/TaskFormReferenceSolutions.tsx
@@ -302,23 +302,18 @@ const TaskFormReferenceSolutions = ({
 
   const onAddReferenceSolution = useCallback(
     () =>
-      setValue(
-        "referenceSolutions",
-        [
-          ...referenceSolutions,
-          {
-            // add unique synthetic id that will be removed later
-            id: Math.max(...referenceSolutions.map((s) => s.id), 0) + 1,
-            isNew: true,
-            isInitial: false,
-            title: "",
-            description: "",
-            tests: [],
-          },
-        ],
-        { shouldDirty: true },
-      ),
-    [referenceSolutions, setValue],
+      setReferenceSolutions([
+        ...referenceSolutions,
+        {
+          id: Math.max(...referenceSolutions.map((s) => s.id), 0) + 1,
+          isNew: true,
+          isInitial: false,
+          title: "",
+          description: "",
+          tests: [],
+        },
+      ]),
+    [referenceSolutions, setReferenceSolutions],
   );
 
   const shouldStopNavigation = useCallback(() => cannotNavigate.current, []);
--- a/frontend/src/components/task/TaskFormReferenceSolutions.tsx
+++ b/frontend/src/components/task/TaskFormReferenceSolutions.tsx
@@ -262,6 +262,7 @@ const TaskFormReferenceSolutions = ({
     watch,
     reset,
     setValue,
+    trigger,
   } = useForm<TaskFormReferenceSolutionsValues & { _fileChanged: boolean }>({
     resolver,
     defaultValues,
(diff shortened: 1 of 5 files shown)


frontend/src/components/task/TaskFormReferenceSolutions.tsx
```diff
@@ -427,7 +427,11 @@ const TaskFormReferenceSolutions = ({
                 data-testid={`solution-${solution.id}`}
               >
                 <div>
-                  <Field.Root>
+                  <Field.Root
```
AthenaPK, line 443: Does this match the way we display validation errors in the rest of the application?

frontend/src/components/task/TaskFormReferenceSolutions.tsx
```diff
@@ -577,13 +570,12 @@ const TaskFormReferenceSolutions = ({
                     ...referenceSolutionFiles,
                     [showSolveTaskModalForId]: solution.file,
                   },
-                  { shouldDirty: true, shouldValidate: true },
+                  { shouldDirty: true },
                 );
```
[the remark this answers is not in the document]
AthenaPK: The robot might have a point here.

frontend/src/components/task/TaskFormReferenceSolutions.tsx
```diff
@@ -292,19 +303,23 @@ const TaskFormReferenceSolutions = ({
...
-          id: Math.max(...referenceSolutions.map((s) => s.id), 0) + 1,
-          isNew: true,
-          isInitial: false,
-          title: "",
-          description: "",
-          tests: [],
-        },
-      ]),
-    [referenceSolutions, setReferenceSolutions],
+      setValue(
```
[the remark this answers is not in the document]
bubudetp: That's the whole point, otherwise empty title and desc flash error descriptions instantly when the row appears.

frontend/src/components/task/TaskFormReferenceSolutions.tsx
```diff
@@ -577,14 +598,15 @@ const TaskFormReferenceSolutions = ({
...
+                  { shouldDirty: true },
                 );
 
                 // react hook form does not detect the file change, so we need to set it manually
                 setValue("_fileChanged", true, {
                   shouldDirty: true,
-                  shouldValidate: true,
                 });
+
+                revalidate();
```
AthenaPK, line 607: We already call this in `updateReferenceSolution` above, do we need to call it twice?

bubudetp: Well, good catch because it should only run once and at the end of the whole operation (`onSave`) because that's when the solution and the solution files are set.

frontend/src/components/task/TaskFormReferenceSolutions.tsx
```diff
@@ -292,19 +302,23 @@ const TaskFormReferenceSolutions = ({
...
-          isInitial: false,
-          title: "",
-          description: "",
-          tests: [],
-        },
-      ]),
-    [referenceSolutions, setReferenceSolutions],
+      setValue(
+        "referenceSolutions",
+        [
```
pierluca: I don't see why we need to call `setValue` instead of just `setReferenceSolutions` here, it seems like an exact copy of the `setReferenceSolutions` logic.

frontend/src/components/task/TaskFormReferenceSolutions.tsx
```diff
@@ -342,46 +352,42 @@ const TaskFormReferenceSolutions = ({
...
+        // and mark the blob as not changed
+        // so the user can navigate without confirmation
+        reset(data);
+
+        toaster.success({
+          id: `task-reference-solutions-save-success-${Date.now()}`,
+          title: intl.formatMessage(messages.saveSuccess),
+          closable: true,
+        });
+      })(e).catch((err) => {
```
pierluca: if we make the `handleSubmit` asynchronous, a `try/catch` syntax might be more coherent than a `catch` ?

pierluca (approved):

PR #597, opened 2026-06-25
[CRT-427] System should track block changes

bubudetp:
## Summary by cubic
Track confirmed Scratch block field changes as student activities and debounce per-keystroke updates to reduce noise for CRT-427. Intermediate keystroke tracking is implemented but disabled until the Scratch upgrade.

- **New Features**
  - Debounce field change tracking (1s) and flush pending changes before non-field events and on unmount.
  - Added `StudentActionType`: `BlockChange`, `IntermediateFieldChange`; map Scratch `change` to `BlockChange` and leave `block_field_intermediate_change` unmapped.
  - Track only when `oldValue !== newValue`; build payload and send via `postStudentAppActivity`.
  - New types: `StudentBlockChangeActivity`, `StudentIntermediateFieldChangeActivity`; extended `WorkspaceChangeEvent` with `group`, `element`, `name`, `oldValue`, `newValue`.
  - Added filters, payload builders, and senders; only `BlockChange` wired into the handler.

- **Bug Fixes**
  - `getChangePayload` returns `StudentBlockChangeActivity` (non-null) for consistent typing.

<sup>Written for commit 7a05559a3f2e3e03b14544a806f350c99bd3142d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/crt25/collimator/pull/597?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/scratch/src/containers/customized-scratch-containers/Blocks.tsx
+++ b/apps/scratch/src/containers/customized-scratch-containers/Blocks.tsx
@@ -213,7 +213,7 @@ class Blocks extends React.Component<Props, State> {
     string,
     { timer: ReturnType<typeof setTimeout>; flush: () => void }
   >();
-  private static readonly FIELD_CHANGE_DEBOUNCE_MS = 500;
+  private static readonly FIELD_CHANGE_DEBOUNCE_MS = 1000;
 
   constructor(props: Props) {
     super(props);
--- a/apps/scratch/src/containers/customized-scratch-containers/Blocks.tsx
+++ b/apps/scratch/src/containers/customized-scratch-containers/Blocks.tsx
@@ -10,7 +10,7 @@ import VM, {
 } from "@scratch/scratch-vm";
 
 import { connect } from "react-redux";
-import ScratchBlocks, { Flyout, Workspace } from "scratch-blocks";
+import ScratchBlocks, { Block, Flyout, Workspace } from "scratch-blocks";
 import { Action, Dispatch } from "redux";
 import log from "@scratch-submodule/packages/scratch-gui/src/lib/log.js";
 import Prompt from "@scratch-submodule/packages/scratch-gui/src/containers/prompt.jsx";
@@ -207,6 +207,12 @@ class Blocks extends React.Component<Props, State> {
   private _renderedToolboxXML?: string;
   private setToolboxRefreshEnabled?: (enabled: boolean) => void;
   private toolboxUpdateTimeout?: number;
+
+  private fieldChangeDebouncers = new Map<
+    string,
+    { timer: ReturnType<typeof setTimeout>; flush: () => void }
+  >();
+  private static readonly FIELD_CHANGE_DEBOUNCE_MS = 500;
 
   constructor(props: Props) {
     super(props);
@@ -461,6 +467,7 @@ class Blocks extends React.Component<Props, State> {
   }
 
   componentWillUnmount() {
+    this.flushPendingFieldChanges();
     this.detachVM();
     this.getWorkspace().dispose();
     clearTimeout(this.toolboxUpdateTimeout);
@@ -1149,17 +1156,69 @@ class Blocks extends React.Component<Props, State> {
       return;
     }
 
+    if (
+      eventAction === StudentActionType.BlockChange &&
+      event.element === "field" &&
+      block
+    ) {
+      this.scheduleFieldChangeTracking(event, block);
+      return;
+    }
+
+    this.trackActivity(event, eventAction, block);
+  }
+
+  private trackActivity(
+    event: WorkspaceChangeEvent,
+    action: StudentActionType,
+    block: Block | null,
+  ) {
     const json = this.props.vm.toJSON();
     const solution = new Blob([json], { type: "application/json" });
 
     handleStudentActivityTracking({
       event,
-      action: eventAction,
+      action,
       canEditTask: this.props.canEditTask,
       sendRequest: this.props.sendRequest,
       solution,
       block,
     });
+  }
+
+  private scheduleFieldChangeTracking(
+    event: WorkspaceChangeEvent,
+    block: Block,
+  ) {
+    const key = `${event.blockId}:${event.name ?? ""}`;
+    const existing = this.fieldChangeDebouncers.get(key);
+
+    if (existing) {
+      clearTimeout(existing.timer);
+    }
+
+    const flush = () => {
+      this.fieldChangeDebouncers.delete(key);
+      this.trackActivity(event, StudentActionType.BlockChange, block);
+    };
+
+    const timer = setTimeout(flush, Blocks.FIELD_CHANGE_DEBOUNCE_MS);
+
+    this.fieldChangeDebouncers.set(key, { timer, flush });
+  }
+
+  private flushPendingFieldChanges() {
+    const pending = Array.from(this.fieldChangeDebouncers.values());
+    this.fieldChangeDebouncers.clear();
+
+    for (const { timer, flush } of pending) {
+      clearTimeout(timer);
+      try {
+        flush();
+      } catch (error) {
+        console.error("Error flushing pending field change activity:", error);
+      }
+    }
   }
 
   render() {
--- a/apps/scratch/src/containers/customized-scratch-containers/Blocks.tsx
+++ b/apps/scratch/src/containers/customized-scratch-containers/Blocks.tsx
@@ -75,6 +75,7 @@ import { getCrtColorsTheme } from "../../blocks/colors";
 import { StudentActionType } from "../../types/scratch-student-activities";
 import {
   handleStudentActivityTracking,
+  isFieldChangeEvent,
   mapScratchEventTypeToStudentActionType,
 } from "../../utilities/scratch-student-activities/student-activity-tracking";
 import { handleBlockLifecycle } from "../../utilities/scratch-student-activities/scratch-block";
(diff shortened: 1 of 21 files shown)


apps/scratch/src/utilities/scratch-student-activities/student-activity-tracking.ts
```diff
@@ -1,12 +1,20 @@
...
+  // trackIntermediateChangeActivity,
+  trackMoveActivity,
+} from ".";
 
 const scratchToStudentActionType: Record<string, StudentActionType> = {
   create: StudentActionType.Create,
   move: StudentActionType.Move,
   delete: StudentActionType.Delete,
+  change: StudentActionType.ConfirmedChange,
+  block_field_intermediate_change: StudentActionType.IntermediateChange,
```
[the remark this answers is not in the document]
bubudetp: 'this path crashes' is a bold statement

apps/scratch/src/types/scratch-student-activities/common.ts
```diff
@@ -26,6 +26,8 @@ export enum StudentActionType {
   Move = "move",
   Delete = "delete",
   GreenFlag = "greenFlag",
+  ConfirmedChange = "confirmedChange",
+  IntermediateChange = "intermediateChange",
```
pierluca: not sure I understand the difference fully from reading this.

Is `ConfirmedChange` any change, or specifically a field that was changed? If it's the latter, I'd consider calling it `FieldChange`.

Is `IntermediateChange` just a keystroke in the fields? If so would `FieldKeystroke` be a better name? Maybe there's a better name, I'd be very open to suggestions.

Perhaps a few comments to explain this would also be helpful.

apps/scratch/src/utilities/scratch-student-activities/student-activity-tracking.ts
```diff
@@ -78,6 +86,38 @@ export const handleStudentActivityTracking = ({
...
+    //   trackIntermediateChangeActivity({
+    //     block,
+    //     sendRequest,
+    //     solution,
+    //     event,
+    //     canEditTask,
+    //   });
+
+    //   break;
+    // }
```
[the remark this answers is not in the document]
pierluca: I don't disagree with the bot and this suggestion.

apps/scratch/src/utilities/scratch-student-activities/student-intermediate-change.ts
pierluca, line 1: as per the other comment, the `intermediate change` vocabulary is a little confusing. if my suggestions are off, maybe @AthenaPK could use her background to help us with naming? :-)

apps/scratch/src/types/scratch-student-activities/common.ts
bubudetp: Two things, sorry for not including this earlier.

First, right now ConfirmedChange is emitted on every keystroke instead of after the change is committed.

Second, IntermediateChange is the new event Scratch introduced for per-keystroke edits. I personally think IntermediateFieldChange is a better name. It makes the field-change intent clear while staying close to Scratch's own terminology.

apps/scratch/src/utilities/scratch-student-activities/student-activity-tracking.ts
pierluca, line 116: if this is emitted per keystroke, shouldn't we debounce to avoid spamming the database on every keystroke?

apps/scratch/src/utilities/scratch-student-activities/student-activity-tracking.ts
```diff
@@ -13,8 +13,19 @@ const scratchToStudentActionType: Record<string, StudentActionType> = {
   create: StudentActionType.Create,
   move: StudentActionType.Move,
   delete: StudentActionType.Delete,
-  change: StudentActionType.ConfirmedChange,
-  // block_field_intermediate_change: StudentActionType.IntermediateChange,
+  /**
+   * Note:
+   *
+   * In the Scratch version we currently have, 'change' is emitted per keystroke
+   * while editing a field, not just on commit. So it already covers the
```
pierluca: "In the Scratch version we currently have" will be outdated very quickly, as a statement. Please use the specific version, or refer to its commit hash.

apps/scratch/src/utilities/scratch-student-activities/student-activity-tracking.ts
```diff
@@ -13,8 +13,19 @@ const scratchToStudentActionType: Record<string, StudentActionType> = {
...
+   * keystroke sequence, which is why intermediate tracking is left off
+   *
+   * A newer Scratch version adds 'block_field_intermediate_change' for
+   * the per-keystroke edits and makes `change` commit-only. Once we upgrade,
+   * uncomment it.
+   */
+  change: StudentActionType.BlockChange,
+  // block_field_intermediate_change: StudentActionType.IntermediateFieldChange,
```
pierluca, line 30: If the event we currently detect maps to the (future Scratch version's) `block_field_intermediate_change`, shouldn't we log `change` as a `StudentActionType.IntermediateFieldChange` for the time being, so that after-the-fact we don't need to distinguish between this and the next Scratch version?

apps/scratch/src/types/scratch-student-activities/common.ts
```diff
@@ -26,8 +26,8 @@ export enum StudentActionType {
   Move = "move",
   Delete = "delete",
   GreenFlag = "greenFlag",
-  ConfirmedChange = "confirmedChange",
-  IntermediateChange = "intermediateChange",
+  BlockChange = "blockChange",
```
pierluca, line 29: Is it a block change or a field change? I'm still a bit confused.

pierluca (changes requested):

apps/scratch/src/types/scratch-student-activities/common.ts
bubudetp: It's not only a field change, but also fires on mutation and checkbox toggles.

apps/scratch/src/utilities/scratch-student-activities/student-activity-tracking.ts
bubudetp: I don't think it's a good idea because the current change event is not equivalent to the future block field intermediate change. I verified in both the current and new scratch-vm blocks.js and change fires for field, mutation, and checkbox while block_field_intermediate_change is field only. In the new version, mutation and checkbox also still arrive as change and not as intermediate events.

So logging all change events as intermediateFieldChange would mislabel mutation and checkbox changes as field edits, which is wrong now and stays wrong after the upgrade.

apps/scratch/src/containers/customized-scratch-containers/Blocks.tsx
```diff
@@ -1149,19 +1161,67 @@ class Blocks extends React.Component<Props, State> {
...
+  private flushPendingFieldChanges() {
+    const pending = Array.from(this.fieldChangeDebouncers.values());
+    this.fieldChangeDebouncers.clear();
+
+    for (const { timer, flush } of pending) {
+      clearTimeout(timer);
+      try {
+        flush();
+      } catch (error) {
+        console.error("Error flushing pending field change activity:", error);
```
pierluca, line 1220: I might be mistaken but I seem to recall we "namespaced" console calls with the current module. 

e.g. https://github.com/crt25/collimator/blob/437b39b80d46f854e74f0b8a99145a8ddcb4c5cd/apps/scratch/src/hooks/useEmbeddedScratch.ts#L195

I realize now that we've forgotten to do this in a bunch of places, perhaps it might be good to have a separate PR addressing this everywhere?

The idea was to make the debugging easier, as we'd know immediately where an error is coming from.

apps/scratch/src/containers/customized-scratch-containers/Blocks.tsx
```diff
@@ -1149,19 +1161,67 @@ class Blocks extends React.Component<Props, State> {
...
+      this.fieldChangeDebouncers.delete(key);
+      this.trackActivity(event, StudentActionType.BlockChange, block);
+    };
+
+    const timer = setTimeout(flush, Blocks.FIELD_CHANGE_DEBOUNCE_MS);
+
+    this.fieldChangeDebouncers.set(key, { timer, flush });
+  }
+
+  private flushPendingFieldChanges() {
```
pierluca, line 1211: The flushing mechanism is smart. Good job 👍

apps/scratch/src/containers/customized-scratch-containers/Blocks.tsx
```diff
@@ -208,6 +209,12 @@ class Blocks extends React.Component<Props, State> {
   private setToolboxRefreshEnabled?: (enabled: boolean) => void;
   private toolboxUpdateTimeout?: number;
 
+  private fieldChangeDebouncers = new Map<
+    string,
+    { timer: ReturnType<typeof setTimeout>; flush: () => void }
+  >();
+  private static readonly FIELD_CHANGE_DEBOUNCE_MS = 500;
```
pierluca: I'm tempted to make this even 1s, so as to minimize the amount of spam to the backend. The flush-pending would catch up anyways if we are too slow.

pierluca (commented):
Are there events that could cause a flush pending more frequently than we want to? have you been able to check that we don't have "noisy" events that would render the debounce moot?

Mostly LGTM at this point!

pierluca (approved):

PR #599, opened 2026-06-29
Add missing namespaces to app logs

bubudetp:
## Summary by cubic
Standardized console log namespaces across Jupyter Notebook Runner and Scratch using file-path–based tags to make logs easier to filter and trace. Supports CRT-427 by making student block-change logs precise, with no runtime changes.

- **Refactors**
  - Adopted file-path–based `logModule` prefixes across autosave, commands (assign/grade/run-all), packages, state, utils, VM loader, GUI/Blocks, and iframe APIs (e.g., “[Jupyter][commands/grade]”, “[Scratch][blocks/helpers]”, “[Jupyter][iframe-api]”).
  - Introduced shared `logBaseModule` for Scratch student activity and composed per-file namespaces in lifecycle and `send-*` modules.
  - Standardized error/warn/debug wording; normalized embedded labels to “[Jupyter][iframe-api]” and “[Scratch][Embedded]”.

<sup>Written for commit ff5a1b24190404416ddc8ace6979a9d804beb2aa. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/crt25/collimator/pull/599?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/jupyter/extensions/notebook-runner/src/commands/assign.ts
+++ b/apps/jupyter/extensions/notebook-runner/src/commands/assign.ts
@@ -9,6 +9,8 @@ import { CannotReadNotebookException } from "../errors/otter-errors";
 import { DEBUG_NOTEBOOK_RUNNER } from "../constants";
 import { copyRequiredFoldersToKernel, handleOtterCommandError } from "./helper";
 
+const logModule = "[Jupyter][Assign]";
+
 export const registerAssignCommand = (
   state: NotebookRunnerState,
   app: JupyterFrontEnd,
@@ -20,7 +22,7 @@ export const registerAssignCommand = (
     isVisible: () => DEBUG_NOTEBOOK_RUNNER,
     execute: async () => {
       try {
-        console.debug("Saving all open notebooks...");
+        console.debug(`${logModule} Saving all open notebooks...`);
         const widgets = app.shell.widgets("main");
 
         const savePromises: Promise<void>[] = [];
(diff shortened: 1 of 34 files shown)


apps/jupyter/extensions/notebook-runner/src/commands/assign.ts
```diff
@@ -9,6 +9,8 @@ import { CannotReadNotebookException } from "../errors/otter-errors";
 import { DEBUG_NOTEBOOK_RUNNER } from "../constants";
 import { copyRequiredFoldersToKernel, handleOtterCommandError } from "./helper";
 
+const logModule = "[Jupyter][Assign]";
```
AthenaPK: Would it make more sense to use the file path to name the log module to help us find the source of the logs more easily, e.g. :
```suggestion
const logModule = "[Jupyter][commands/assign]"
```

apps/scratch/src/utilities/scratch-student-activities/log-module.ts
AthenaPK, line 1: This makes me wonder, what level of granularity should we go for when assigning log module names in different files? (I see that in jupyter we use one label per file, for example).
I think it makes sense to do that here and perhaps we could also apply this logic in some places in the jupyter logs (e.g. some files in the notebook-runner directory, `loading-state.ts` probably doesn't need a separate log label), what do you think?

AthenaPK (commented):
Overall LGTM, just a couple small comments : )
Also, do we have a ticket associated with this? If not, it might be a good idea to create one, so that we keep track of the work.

apps/scratch/src/utilities/scratch-student-activities/log-module.ts
pierluca: So far (elsewhere) the idea was to have 1 file = 1 `logModule`, so as to narrow down the cause pretty quickly, but it does seem to make sense to have a shared logModule in some areas.

My main concern is that `{logModule} Could not find xml in event` (with the same `logModule`) can now show up in multiple files, which brings ambiguity back...

AthenaPK (approved):

PR #603, opened 2026-07-02
[CRT-434] Student progress is not reported consistently

bubudetp:
## Summary by cubic
Fix inconsistent student progress across Dashboard and Task Instance (CRT-434) by centralizing status calculation and using current analyses to show in‑progress work. Status text and colors now match across screens.

- **Bug Fixes**
  - Use shared `getTaskStatus` to compute status from `ExistingStudentSolution` or `CurrentStudentAnalysis` (not started, incomplete, complete).
  - Populate `currentAnalysis` in Dashboard via `useAllSessionCurrentAnalyses` so in‑progress work appears.
  - Use `getTaskStatusColor` for consistent indicators.

- **Refactors**
  - Added `frontend/src/components/task-progress.ts` with `TaskStatus`, `getTaskStatus`, and `getTaskStatusColor`.
  - Removed duplicate status logic from `ProgressList.tsx` and `TaskInstanceProgressList.tsx`.

<sup>Written for commit b8488040d34ed27713a1bfee42432450fd4e9be2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/crt25/collimator/pull/603?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/src/components/dashboard/ProgressList.tsx
+++ b/frontend/src/components/dashboard/ProgressList.tsx
@@ -1,4 +1,4 @@
-import { ComponentProps, useMemo } from "react";
+import { useMemo } from "react";
 import { defineMessages, useIntl, FormattedMessage } from "react-intl";
 import styled from "@emotion/styled";
 import { HStack, Icon, Link, Status, Text } from "@chakra-ui/react";
@@ -13,6 +13,11 @@ import { CurrentStudentAnalysis } from "@/api/collimator/models/solutions/curren
 import { ColumnType } from "@/types/tanstack-types";
 import { ProgressMessages } from "@/i18n/progress-messages";
 import { useAllSessionCurrentAnalyses } from "@/api/collimator/hooks/solutions/useAllSessionCurrentAnalyses";
+import {
+  getTaskStatus,
+  getTaskStatusColor,
+  TaskStatus,
+} from "../task-progress";
 import MultiSwrContent from "../MultiSwrContent";
 import { StudentName } from "../encryption/StudentName";
 import ChakraDataTable, { ColumnSize } from "../ChakraDataTable";
@@ -48,6 +53,7 @@ const messages = defineMessages({
 type TaskSolutions = {
   taskId: number;
   solutions: ExistingStudentSolution[];
+  currentAnalysis: CurrentStudentAnalysis | null;
 };
 
 type AnonymousStudent = {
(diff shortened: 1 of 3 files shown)


pierluca (commented):
It seems that `TaskInstanceProgressList` and `ProgressList` are almost identical, except for the rendering of the table.
- Should they be merged? This would avoid future bugs like this one, where they behave differently.
- Should we define some higher-order hook (e.g. `useStudentProgress()` that does all the pre-processing between solutions and analyses? perhaps even including resolving the status?

frontend/src/components/task-instance/TaskInstanceProgressList.tsx
pierluca, line 130: `TaskTemplate` could be extracted into its own component, such that whatever is now in `task-progress` become helper functions for this component.

I'm aware that `ProgressList` does more with it than `TaskInstanceProgressList`, but the former is a superset of the latter, so this could be addressed in a couple of ways (e.g., parameter in the new component, composite component, etc.)

pierluca, line 102: shouldn't we add `useCallback` here ?

frontend/src/components/dashboard/ProgressList.tsx
pierluca, line 157: currently this is always displayed, and it doesn't provide any value (yet).
Should we just comment out its content for now? 

```suggestion
const helpTemplate = (_rowData: StudentProgress) => (</>); /* (
<Icon>
<LuHand />
</Icon>
); */
```

pierluca (commented):
Given a new PR has been opened for the refactoring, I am okay with keeping the two files separate.

However, there is one more thing that can be merged, highlighted in the comment here, and a couple of other (unrelated) issues I spotted.

Happy to have those addressed in separate PRs and merge one after the other :)

frontend/src/components/task-instance/TaskInstanceProgressList.tsx
bubudetp: why? the nameTemplate is already memoized as part of the column def:
https://github.com/crt25/collimator/blob/main/frontend/src/components/task-instance/TaskInstanceProgressList.tsx#L283-L291

pierluca: Sorry, I completely missed that `nameTemplate` sits outside the component definition.

pierluca (approved):
Happy to approve this given the changes in #605 that will follow

PR #604, opened 2026-07-08
[CRT-436] Authenticated student is registered multiple times when multiple teacher tabs are open during session join

bubudetp:
---
## Summary by cubic
Prevents duplicate student records when multiple teacher tabs are open by deduplicating with a deterministic, class-scoped identifier and making sign-in concurrency-safe. Also fixes socket cleanup and resets the auth lock on failures. [CRT-436]

- **Bug Fixes**
  - Add Base64url `studentIdentifier` to `AuthenticatedStudent` with a partial unique index per class; derived with `classId`; service resolves by identifier first, then pseudonym; backfills legacy rows; retries after unique constraint races.
  - API: `studentIdentifier` is required in `StudentAuthenticationRequestDto`; controllers/services accept it; response remains unchanged (identifier not returned).
  - Frontend: derive identifier via `TeacherLongTermKeyPair.deriveStudentIdentifier()` (HMAC) and include it in auth requests; update generated types.
  - Reliability: release `isAuthenticating` on auth failure in the join flow; set `client.data.user` on socket connect to enable proper disconnection and prevent `socketsByUserId` leaks.

<sup>Written for commit 845244a7c3ec7dbdff49598d9ffc2dfc9839bd8b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/crt25/collimator/pull/604?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

[CRT-436]: https://classroomrt.atlassian.net/browse/CRT-436?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backend/src/api/authentication/dto/student-authentication-response.dto.ts
```diff
@@ -21,6 +21,17 @@ export class StudentAuthenticationResponseDto {
...
+  @IsNotEmpty()
+  @IsOptional()
+  @ApiProperty({
+    required: false,
+    description:
+      "The deterministic, teacher-keyed identifier stored for the student, encoded in Base64url. Absent for students created before this identifier existed.",
+  })
+  @Expose()
+  readonly studentIdentifier?: string;
+
```
pierluca: do we want to return this?

frontend/src/hooks/useHandleStudentAuthenticationRequest.ts
```diff
@@ -127,16 +127,20 @@ export const useHandleStudentAuthenticationRequest = (): CallbackType => {
...
-      );
+      const [pseudonym, studentIdentifier] = await Promise.all([
+        findOrCreatePseudonym(
+          longTermIdentifier,
+          name,
+          klass,
+          authContext.keyPair,
+        ),
+        authContext.keyPair.deriveStudentIdentifier(longTermIdentifier),
+      ]);
```
pierluca, line 141: it seems like the pseudonym, based on the long term identifier, would be a sufficient student identifier. Why do we need the extra `deriveStudentIdentifier` ?
I might be missing something here.

frontend/src/pages/class/[classId]/session/[sessionId]/join.tsx
```diff
@@ -297,6 +297,8 @@ const JoinSession = () => {
           `${logModule} Failed to generate student key pair`,
           error,
         );
+
+        isAuthenticating.current = false;
```
Comment (pierluca) [inline frontend/src/pages/class/[classId]/session/[sessionId]/join.tsx:301]: Not your fault, but this `useEffect` has grown to an insane size and deserves its own refactoring at some point 😱 Any change larger than this and I'd have pushed to do this sooner rather than later.

frontend/src/hooks/useHandleStudentAuthenticationRequest.ts
bubudetp: Pseudonyms are encrypted randomly even with the same input data. When a student joins at the same time where there are two tabs open, the lookup for existing matching pseudonym returns empty, which leads to creating two different pseudonyms.

backend/src/api/authentication/dto/student-authentication-response.dto.ts
bubudetp: No, we already compute that in the frontend, good catch

frontend/src/pages/class/[classId]/session/[sessionId]/join.tsx
Comment (bubudetp) [reply frontend/src/pages/class/[classId]/session/[sessionId]/join.tsx:301]: Yeah agreed, maybe not in this PR though ^^

frontend/src/hooks/useHandleStudentAuthenticationRequest.ts
pierluca: what do you mean encrypted randomly? I don't see what's random about this?
https://github.com/crt25/collimator/blob/8a141f3de4730e9dcb6f0771f1f37f9ec3c8ef05/frontend/src/hooks/useHandleStudentAuthenticationRequest.ts#L53C64-L61C6

bubudetp: see here: https://github.com/crt25/collimator/blob/main/frontend/src/utilities/crypto/SymmetricKey.ts#L42-L75

backend/src/api/authentication/authentication.service.ts
```diff
@@ -380,49 +383,141 @@ export class AuthenticationService {
...
+
+          student = await this.prisma.student.findUniqueOrThrow({
+            where: { id: authenticatedStudent.studentId, deletedAt: null },
+          });
+        } else {
+          throw error;
+        }
+      }
+    } else {
+      student = await this.prisma.student.findUniqueOrThrow({
```
pierluca, line 506: Discussed today with @bubudetp: 
- if studentIdentifier is set but not in database -> update DB
- everywhere else: studentIdentifier becomes required.

pierluca (approved):

PR #607, opened 2026-07-09
[CRT-420][CRT-421] Students results view display that past solutions were starred

bubudetp:
## Summary by cubic
Shows the latest submission as the current version (even if starred) and clearly labels past starred versions. Unifies starring across submissions and activity-only solutions via `SolutionActivityReference`, with batch star/unstar capped at 100 hashes. Addresses CRT-420/CRT-421.

- **Bug Fixes**
  - SQL: emit `isLatest`; derive `isReference` via `SolutionActivityReference`; include activity-only stars; dedupe by solution hash; ignore soft-deleted duplicates.
  - API: replace source-specific PATCH endpoints with `/api/v0/classes/:classId/sessions/:sessionId/task/:taskId/solutions/student/:studentId/reference`; accept base64url `solutionHashes[]` (max 100); expose `isLatest` on `CurrentStudentAnalysisDto`.
  - Service/Authorization: validate hashes belong to class/session/task; star/unstar by hash across sources; batch create/delete transactionally; authorize class owner or admin.
  - Frontend: add `usePatchStudentReferenceSolutions` and use it in `useStarAnalysis`; batch unstar in `UnstarPastSolutionsButton`; pick current via `isLatest`; derive and show past-starred excluding current; update `TaskInstanceProgressList`; regenerate clients/mocks; add tests for batched cache updates.

- **Migration**
  - Create `SolutionActivityReference` with `createdAt`; backfill existing stars; drop `isReference` from `StudentSolution`.
  - Update callers to the new endpoint and DTO (`PatchStudentReferenceSolutionDto`).

<sup>Written for commit 268e6161631317b7c9784952d3a753d9d82a0a5f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/crt25/collimator/pull/607?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/src/api/solutions/solutions.service.ts
pierluca, line 1: won't commit d42a9102c7e77984bd26b3d196d8b5f9084a8f52 lead to seeing the same solution displayed potentially many, many times in the showcase?

frontend/src/components/task-instance/TaskInstanceProgressList.tsx
```diff
@@ -262,20 +257,34 @@ const TaskInstanceProgressList = ({
...
+            .filter(
+              (a): a is CurrentStudentAnalysis =>
+                // the analysis is for this student, is a starred solution and is not the current version
+                a instanceof CurrentStudentAnalysis &&
+                a.studentId === student.studentId &&
+                a.isReferenceSolution &&
+                a.solutionHash !== currentAnalysis?.solutionHash,
+            )
+            .map((a) => [a.solutionHash, a] as const),
+        ).values(),
```
pierluca: if I understand the intention here, we're de-duplicating by solutionHash.
However, `useStarAnalysis` removes `isReference` by `analysis.studentSolutionId` (student solution) and separately by `analysis.solutionHash` (student activity). 

Doesn't this mean that we could have left-over student solutions that are _not_ un-starred if they conflict in terms of solutionHash?

pierluca (changes requested):
Also, the CI fails :)

frontend/src/components/task-instance/TaskInstanceProgressList.tsx
```diff
@@ -262,20 +257,34 @@ const TaskInstanceProgressList = ({
...
+      const pastStarredAnalyses = [
+        ...new Map(
+          (currentAnalyses ?? [])
+            .filter(
+              (a): a is CurrentStudentAnalysis =>
+                // the analysis is for this student, is a starred solution and is not the current version
+                a instanceof CurrentStudentAnalysis &&
+                a.studentId === student.studentId &&
+                a.isReferenceSolution &&
+                a.solutionHash !== currentAnalysis?.solutionHash,
```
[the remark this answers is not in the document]
bubudetp: Arguably, we were always displaying stars on analysed solutions and not on the generalised task solutions. This bug does not apply, because isLatest is always set on the currentAnalysis in when retrieving them.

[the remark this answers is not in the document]
bubudetp: NVM just remembered that an unanalyzed task solution does not produce a solution row, so this is wrong

backend/src/api/solutions/solutions.service.ts
```diff
@@ -248,8 +255,10 @@ export class SolutionsService {
         studentSolutionId: analysis.studentSolutionId,
         isStudentSolution: analysis.isStudentSolution ?? false,
         studentKeyPairId: analysis.studentKeyPairId,
+        isLatest: analysis.isLatest ?? false,
```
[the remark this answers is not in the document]
bubudetp: The solution rows are created only after being analysed so this does not hold

backend/src/api/solutions/solutions.service.ts
bubudetp: yeah the bot caught that too, that's my bad 😅 

I excluded the latest solution from the starred solution branch with AND NOT EXISTS, just learned about it, it's awesome: https://stackoverflow.com/questions/173041/not-in-vs-not-exists

frontend/src/components/task-instance/TaskInstanceProgressList.tsx
bubudetp: hmm we could have left-over student solutions in case we have a student solution and a student activity with the same solution hash, which is likely possible. The other case I can think of is two student solutions but with the same content hash.

I assume we can dedup based on how we unstar so by studentSolutionId for solutions or by solutionHash for activities

backend/src/api/solutions/solutions.service.spec.ts
```diff
@@ -67,7 +69,7 @@ const buildReferenceAnalysisRow = (
   ...overrides,
 });
 
-describe("SolutionsService", () => {
+describe.only("SolutionsService", () => {
```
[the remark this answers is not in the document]
pierluca: Major issue here :)

frontend/src/components/task-instance/TaskInstanceProgressList.tsx
```diff
@@ -262,20 +257,42 @@ const TaskInstanceProgressList = ({
...
+            )
+            .map(
+              (a) =>
+                [
+                  a.isStudentSolution
+                    ? `solution:${a.studentSolutionId}`
+                    : `activity:${a.solutionHash}`,
+                  a,
+                ] as const,
+            ),
```
pierluca: not sure I fully follow, but it also seems like this won't ever delete anything, given the backend is not aware of this? 🤔 

In general I wouldn't rely on string processing to build something robust and type-checkable: if we need two collections, we can have two arrays.

frontend/src/components/task-instance/TaskInstanceProgressList.tsx
pierluca: I hadn't thought of the two students with the same solution, but it seems the `baseWhere` takes care of that in the backend. 

Copying `useStarAnalysis` here for clarity:

```ts

export const useStarAnalysis = (): StarAnalysisType => {
  const patchStudentSolutionIsReference = usePatchStudentSolutionIsReference();
  const patchStudentActivityIsReference = usePatchStudentActivityIsReference();

  return useCallback<StarAnalysisType>(
    (classId, analysis, isReference) => {
      if (analysis.isStudentSolution) {

        // !! this can unstar a specific solution but miss others
        return patchStudentSolutionIsReference(
          classId,
          analysis.sessionId,
          analysis.taskId,
          analysis.studentSolutionId!,
          { isReference },
        );
      }

      // !! this can unstar the latest solution, if it shares the same solution hash
      return patchStudentActivityIsReference(
        classId,
        analysis.sessionId,
        analysis.taskId,
        analysis.studentId,
        analysis.solutionHash,
        isReference,
      );
```

essentially, why would we de-duplicate the analyses we want to unstar? 🤔

pierluca (changes requested):

frontend/src/components/solution/UnstarPastSolutionsButton.tsx
```diff
@@ -43,48 +43,31 @@ const messages = defineMessages({
...
-      await Promise.all(
-        starredPastSolutions.map((s) =>
-          patchSolution(classId, sessionId, taskId, s.id, {
-            isReference: false,
-          }),
-        ),
-      );
+      for (const analysis of pastStarredAnalyses) {
+        await starAnalysis(classId, analysis, false);
+      }
```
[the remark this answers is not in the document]
pierluca: This, and this looks like a client-mediated N+1.
It might make sense to be able to batch-send everything that needs to be unstarred

backend/prisma/schema.prisma
```diff
@@ -374,3 +377,22 @@ model StudentActivityApp {
...
+  studentId    Int
+  sessionId    Int
+  classId      Int
+  taskId       Int
+
+  solution Solution @relation(fields: [taskId, solutionHash], references: [taskId, hash], onDelete: Restrict)
+  student  Student  @relation(fields: [studentId], references: [id], onDelete: Cascade)
+  session  Session  @relation(fields: [sessionId], references: [id], onDelete: Restrict)
+  class    Class    @relation(fields: [classId], references: [id], onDelete: Restrict)
+  task     Task     @relation(fields: [taskId], references: [id], onDelete: Restrict)
```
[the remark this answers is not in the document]
bubudetp: What do you mean by referenced session

backend/prisma/schema.prisma
```diff
@@ -380,21 +379,18 @@ model StudentActivityApp {
 }
 
 model SolutionActivityReference {
-  id           Int      @id @default(autoincrement())
-  createdAt    DateTime @default(now())
```
pierluca: for debugging purposes once it's in production, it might be nice to keep the `createdAt` column.

backend/src/api/solutions/dto/existing-student-solution.dto.ts
```diff
@@ -32,15 +32,6 @@ export class ExistingStudentSolutionDto
...
   readonly createdAt!: Date;
 
-  @ApiProperty({
-    example: "true",
-    description: "Whether this solution is marked as a reference solution.",
-    type: "boolean",
-  })
-  @Type(() => Boolean)
-  @Expose()
-  readonly isReference!: boolean;
```
pierluca, line 50: are you sure about removing this? 
I'm thinking that from an API perspective it might be nice to have this extra bit of information.

backend/src/api/solutions/solutions.service.ts
```diff
@@ -49,7 +50,19 @@ type WithTestsAndSolution<T> = T & {
...
+> & {
+  solution: SolutionWithoutData & {
+    activityReferences: SolutionActivityReference[];
+  };
+};
+
+export type StudentSolutionWithoutData =
+  WithTestsAndSolution<StudentSolution> & {
+    isReference: boolean;
+  };
```
pierluca: I'm a bit confused by these types.
The alterations with no clear comments make me wonder if we're missing something somewhere else.

backend/src/api/solutions/solutions.service.ts
```diff
@@ -49,7 +50,19 @@ type WithTestsAndSolution<T> = T & {
   solution: SolutionWithoutData;
 };
 
-export type StudentSolutionWithoutData = WithTestsAndSolution<StudentSolution>;
+type StudentSolutionWithReferences = Omit<
+  WithTestsAndSolution<StudentSolution>,
+  "solution"
+> & {
+  solution: SolutionWithoutData & {
+    activityReferences: SolutionActivityReference[];
```
pierluca: why would we want to return all the references at once, with just one solution? This seems odd to me

backend/src/api/solutions/solutions.service.ts
```diff
@@ -115,25 +129,52 @@ export class SolutionsService {
...
+          tests: {
+            where: includeSoftDelete ? {} : { deletedAt: null },
+          },
+          solution: {
+            omit: omitData,
+            include: {
+              activityReferences: {
+                where: { sessionId, taskId },
+              },
+            },
```
pierluca: I'm not sure why we need to do this `include` followed by the `WithReferenceFlag` for this.

Could we simply do something along the lines of 
```
        _count: {
          select: {
            activityReferences: {
              where: { sessionId, taskId },
            },
          },
        },
```

and cast the _count to boolean afterwards?

backend/src/api/solutions/solutions.service.ts
```diff
@@ -455,85 +506,77 @@ export class SolutionsService {
...
+    const studentSolutions = await this.prisma.studentSolution.findMany({
       ...args,
       where,
       include: {
         solution: {
           omit: omitData,
+          include: {
+            activityReferences: true,
+          },
         },
```
pierluca: same question

backend/src/api/solutions/solutions.service.ts
```diff
@@ -619,7 +662,13 @@ export class SolutionsService {
     const studentSolution = await this.prisma.studentSolution.create({
       data: checkedStudentSolution,
       include: {
-        solution: true,
+        solution: {
+          include: {
+            activityReferences: {
+              where: { studentId, sessionId, taskId },
+            },
+          },
```
pierluca: same again :)

pierluca (commented):
I tried not to be too picky this time around.

Given the project stage, I'm a bit concerned by the scope of this, but it seems in a reasonably good state.

I just have a few questions, but I think it's close to being in a mergeable state.

frontend/src/components/task-instance/TaskInstanceProgressList.tsx
pierluca: Claude Review: this concern was real against the design as it stood (unstar-by-`studentSolutionId` for solutions vs unstar-by-`solutionHash` for activities could indeed strand a star when the same hash existed in both sources) — but the final architecture resolves it structurally rather than by dedup logic. `SolutionActivityReference` is keyed by `(studentId, sessionId, taskId, solutionHash)` only — no source row id — so a StudentSolution and a StudentActivity sharing a hash map to **one** reference row, and both the star and unstar paths (`useStarAnalysis` → the single `patchStudentReferenceSolutions` call; the dual per-source hooks are deleted) read and write that same row. Both SQL branches derive `isReference` from a LEFT JOIN to this same table on the identical key. The leftover-star scenario cannot occur at head (`f57b0861`); this thread can be resolved as superseded by the `SolutionActivityReference` migration. — posted by Claude during an automated adversarial review.

backend/src/api/solutions/dto/existing-student-solution.dto.ts
bubudetp: We don't use it anywhere in the codebase though

backend/src/api/solutions/solutions.service.ts
bubudetp: This is outdated..

backend/src/api/solutions/solutions.service.ts
bubudetp: same as above..

backend/src/api/solutions/solutions.service.ts
bubudetp: same as above

backend/src/api/solutions/solutions.service.ts
bubudetp: same as above

backend/src/api/solutions/solutions.service.ts
bubudetp: same as above

pierluca (approved):

backend/prisma/schema.prisma
```diff
@@ -342,3 +345,21 @@ model StudentActivityApp {
 
   studentActivity StudentActivity @relation(fields: [id], references: [id], onDelete: Cascade)
 }
+
+model SolutionActivityReference {
+  createdAt DateTime @default(now())
```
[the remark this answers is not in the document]
bubudetp: kept for debugging principles in prod environment

backend/prisma/migrations/20260730083645_add_created_at_to_solution_activity_reference/migration.sql
```diff
@@ -0,0 +1,2 @@
+-- AlterTable
+ALTER TABLE "SolutionActivityReference" ADD COLUMN     "createdAt" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP;
```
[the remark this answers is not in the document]
bubudetp: there weren't any previously, so there's no need

PR #608, opened 2026-07-16
[CRT-388] Refresh block limit indicator when blocks are dragged to another sprite

bubudetp:
## Summary by cubic
Refreshes the block limit indicator when blocks are dragged onto another sprite so counts stay accurate and the limit UI updates as expected (CRT-388).

- Bug Fixes
  - Listens for endDrag and defers the refresh until the VM copies the block to the target, covering cases without create/delete events.
  - Prevents false limit hits by tracking the block created in the current drag and excluding its id from limit checks when dropping to another sprite; clears this tracking at endDrag and after cross-sprite drops.
  - Parses opcodes from cloned event XML to update buttons, removes the temporary helper element, and only runs when dropped outside and XML exists.

- Refactors
  - Extracted `refreshBlockConfigButtonsFromXml`, imported `updateSingleBlockConfigButton` directly, removed leftover lifecycle params/call sites, and deleted outdated ESLint disable comments.

<sup>Written for commit db4aa7e036bd8e95e7382d09a4bb63de12c58abf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/crt25/collimator/pull/608?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/scratch/src/blocks/helpers.ts
+++ b/apps/scratch/src/blocks/helpers.ts
@@ -13,6 +13,10 @@ export const ignoreEvent = (event: MouseEvent): void => {
 
 const pendingRejectionBlockIds = new Set<string>();
 
+// Blocks created during the current drag gesture
+// Used to detect when a block dragged onto another sprite was created in the same gesture rather
+const blocksCreatedInCurrentGesture = new Set<string>();
+
 const shouldPreventBlockCreation = (
   event: WorkspaceChangeEvent,
   vm: VMExtended,
@@ -34,6 +38,9 @@ const shouldPreventBlockCreation = (
   if (isEndDrag) {
     const blockId = event.blockId ?? "";
 
+    // the drag gesture has ended, forget any blocks created during it
+    blocksCreatedInCurrentGesture.clear();
+
     if (pendingRejectionBlockIds.has(blockId)) {
       pendingRejectionBlockIds.delete(blockId);
       workspace.undo(false);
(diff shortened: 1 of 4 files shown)


apps/scratch/src/utilities/scratch-student-activities/scratch-block/lifecycle.ts
```diff
@@ -57,6 +57,34 @@ export const handleBlockLifecycle = ({
...
+        // shareBlocksToTarget resolves in a promise callback), so no
+        // create/delete event fires on the active workspace. Defer the
+        // counter refresh until the copy is registered in the runtime.
+        if (!event.isOutside || !event.xml) {
+          return;
+        }
+
+        // Clone: the dragged block's xml element may still be referenced by
+        // scratch-blocks, so it must not be re-parented like create/delete xml
+        const el = document.createElement("div");
```
pierluca: should this be removed afterwards?

apps/scratch/src/utilities/scratch-student-activities/scratch-block/lifecycle.ts
```diff
@@ -57,6 +57,34 @@ export const handleBlockLifecycle = ({
...
+        el.appendChild(event.xml.cloneNode(true));
+
+        const opcodes = [...el.querySelectorAll("block[type]")]
+          .map((element) => element.getAttribute("type"))
+          .filter(filterNonNull);
+
+        setTimeout(() => {
```
pierluca: is a set timeout without delay sufficient?

apps/scratch/src/blocks/helpers.ts
```diff
@@ -53,10 +60,29 @@ const shouldPreventBlockCreation = (
...
+  const eventBlockId = event.blockId ?? "";
+
+  const excludeBlockId =
+    isBlockDraggedToSprite && blocksCreatedInCurrentGesture.has(eventBlockId)
+      ? event.blockId
+      : undefined;
+
+  if (!wouldExceedLimits(vm, block, excludeBlockId)) {
+    if (isBlockCreated) {
+      blocksCreatedInCurrentGesture.add(eventBlockId);
```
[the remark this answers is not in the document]
bubudetp: every drag gesture terminates with an endDrag and the set is emptied at the end of every completed gesture --> an existing block from an earlier gesture cannot remain in the set

apps/scratch/src/blocks/helpers.ts
```diff
@@ -13,6 +13,10 @@ export const ignoreEvent = (event: MouseEvent): void => {
 
 const pendingRejectionBlockIds = new Set<string>();
 
+// Blocks created during the current drag gesture
+// Used to detect when a block dragged onto another sprite was created in the same gesture rather
+const blocksCreatedInCurrentGesture = new Set<string>();
```
pierluca: I'll need to review more, but why should this be a set as opposed to a single optional element?

bubudetp: I thought a gesture could create multiple blocks, I just tested every block in our toolbox and was not able to see 2 creation at once so it indeed can only be an optional element.

pierluca (approved):
Variables may need some renaming in a subsequent PR:)

PR #612, opened 2026-07-19
[CRT-438] Freeze Jupyter task editor read-only during save

pierluca:
## Problem (CRT-438)

While a teacher runs the task save/export pipeline (`getTask`), the notebook must not be edited: the student copy is generated up front by the assign pipeline, but the teacher template is read at the end, so an edit in between lands only in the template and desyncs the two.

### Why the first approach was wrong (reported broken)

The initial version set `cell.readOnly = true` on every cell. In JupyterLab 4.4 the `Notebook` widget sets `syncEditable = true` on its cells, so the `readOnly` setter writes **`editable: false` into the cell metadata**, which is serialized into the saved `.ipynb`. The pipeline saves the notebook while frozen, so `editable: false` got baked into **both** the teacher template and the generated student copy — students could no longer type, and reopening the task still showed the cells locked.

## Fix

Block the **whole shell** for the duration of `getTask` instead of touching any cell:

- `app.shell.node.inert = true` — disables user pointer, keyboard and focus across the JupyterLab UI, while leaving programmatic work (command execution, kernel messaging, `context.save()`) unaffected. No model/metadata is touched, so nothing is persisted.
- a fixed-position spinner overlay (`jp-Spinner`) for visual feedback.
- both torn down in `finally` (idempotent cleanup), so the error path can't strand the UI.

`inert` is scoped to the shell node, so the otter error dialogs (`showErrorMessage`, which attach to `document.body` and are fire-and-forget) remain visible and dismissible.

New helper: `src/ui-blocker.ts`. `getTask` reverts to its original body wrapped in `blockUserInterface(this.app)` / `unblock()`.

## ⚠️ Data note

Any task **saved while the old `readOnly` approach was deployed** has `editable: false` baked into its notebook cells and will stay locked until re-saved with this fix. If any such tasks exist, they need a re-save (or a one-off metadata strip on import).

## Verification

Extension `eslint` + `tsc --noEmit` clean (0 errors). Needs an in-app check on a rebuilt JupyterLite: run a task save → confirm the UI is fully inert with a spinner during the pipeline, editable again afterward, and that the saved student/teacher notebooks are **not** read-only on reload.

### DIFF
--- a/apps/jupyter/extensions/notebook-runner/src/ui-blocker.ts
+++ b/apps/jupyter/extensions/notebook-runner/src/ui-blocker.ts
@@ -1,5 +1,7 @@
 import { JupyterFrontEnd } from "@jupyterlab/application";
 import { Spinner } from "@jupyterlab/apputils";
+import { messages } from "./i18n/messages";
+import { formatMessage } from "./i18n/intl";
 
 const OVERLAY_ID = "notebook-runner-ui-blocker";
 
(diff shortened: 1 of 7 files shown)


apps/jupyter/extensions/notebook-runner/src/ui-blocker.ts
```diff
@@ -0,0 +1,63 @@
...
+  overlay.appendChild(label);
+
+  document.body.appendChild(overlay);
+
+  let cleanedUp = false;
+  return (): void => {
+    if (cleanedUp) {
+      return;
+    }
+    cleanedUp = true;
```
pierluca, line 59:
```suggestion

    cleanedUp = true;
```

bubudetp (approved):

PR #615, opened 2026-07-19
[CRT-399] Auto-select the Pyodide kernel instead of prompting

pierluca:
## CRT-399 — Auto-select the Pyodide kernel instead of prompting

Opening a task notebook popped a **"Select the kernel"** dialog for task creators (it should auto-select Python (Pyodide) silently).

### Root cause — a kernelspec-readiness race
The visible task notebook opens (and its `sessionContext` resolves a kernel) **before the Pyodide kernelspec is registered**. `SessionContext.getDefaultKernel` then sees `specs === null`, so `startKernel()` returns `true` and JupyterLab falls back to the kernel-select dialog. Once the specs load, the single Python-language kernel resolves automatically — so the bug is purely ordering.

### Fix
`await app.serviceManager.kernelspecs.ready` (+ a non-empty `specsChanged` guard) **before** opening the visible notebook — all five `openOrReveal` sites in `iframe-api.ts` now route through a small `openTaskNotebook()` helper. This lets `getDefaultKernel` deterministically resolve the single Pyodide kernel with no prompt, and fixes the same race for the student/solve open too.

Also removed the pre-existing `notebookTracker.widgetAdded` `autoStartDefault: true` hook — it **cannot** fix this (the `!specs` branch short-circuits before `autoStartDefault` is read, and `NotebookPanel.setConfig` re-applies the setting's `false` default anyway), so it was dead and misleading.

### Verification
- Extension `tsc --noEmit`: **zero errors from the change** (remaining errors are the pre-existing uninitialized `./iframe-rpc` submodule, on `main` too); the `kernelspecs.ready/.specs/.specsChanged` APIs typecheck.
- ⚠️ **Not verified in the running app** (needs the built JupyterLite/pyodide). Please confirm: `?mode=edit` opening `task.ipynb` shows **no** kernel dialog and auto-selects Python (Pyodide); same in `?mode=solve`.
- Confidence HIGH on mechanism + fix; the exact race window is the one part worth an in-app sanity check.

> ⚠️ Draft, opened autonomously.

---
## Summary by cubic
Auto-selects the Python (Pyodide) kernel by waiting for kernelspecs before opening task notebooks, removing the “Select the kernel” prompt in edit and solve modes and fixing CRT-399. The wait is bounded to 15s with a safe fallback, and a lint rule prevents regressions.

- **Bug Fixes**
  - Added `waitForKernelSpecs()` that awaits `serviceManager.kernelspecs.ready`, listens to `specsChanged` and checks the emitted specs payload, re-checks immediately after connecting, and on a 15s timeout opens the notebook anyway (dialog may appear).
  - Routed all notebook opens through `openTaskNotebook()`; lint-banned direct `documentManager.openOrReveal` calls.
  - Removed the `notebookTracker.widgetAdded` `autoStartDefault` hook, which didn’t affect the missing-specs case.

<sup>Written for commit cec5d928e999d59aa2a7d403c01b9aeb5bb1e5ba. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/crt25/collimator/pull/615?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/jupyter/extensions/notebook-runner/src/iframe-api.ts
+++ b/apps/jupyter/extensions/notebook-runner/src/iframe-api.ts
@@ -140,17 +140,21 @@ export class EmbeddedPythonCallbacks {
     const kernelSpecs = this.app.serviceManager.kernelspecs;
     await kernelSpecs.ready;
 
-    const hasSpec = (): boolean =>
-      !!kernelSpecs.specs &&
-      Object.keys(kernelSpecs.specs.kernelspecs).length > 0;
+    const hasSpec = (specs = kernelSpecs.specs): boolean =>
+      !!specs && Object.keys(specs.kernelspecs).length > 0;
 
     if (hasSpec()) {
       return;
     }
 
     await new Promise<void>((resolve) => {
-      function onChange(): void {
-        if (hasSpec()) {
+      // specsChanged emits the freshly fetched specs, so use the emitted
+      // payload instead of re-reading the manager's mutable specs property
+      function onChange(
+        _sender: unknown,
+        specs: typeof kernelSpecs.specs,
+      ): void {
+        if (hasSpec(specs)) {
           clearTimeout(timeout);
           kernelSpecs.specsChanged.disconnect(onChange);
           resolve();
--- a/apps/jupyter/extensions/notebook-runner/src/iframe-api.ts
+++ b/apps/jupyter/extensions/notebook-runner/src/iframe-api.ts
@@ -125,6 +125,43 @@ export class EmbeddedPythonCallbacks {
       ? EmbeddedPythonCallbacks.taskTemplateLocation
       : EmbeddedPythonCallbacks.studentTaskLocation;
 
+  /**
+   * Resolves once at least one kernelspec (the Pyodide kernel) is registered.
+   */
+  private async waitForKernelSpecs(): Promise<void> {
+    const kernelSpecs = this.app.serviceManager.kernelspecs;
+    await kernelSpecs.ready;
+
+    const hasSpec = (): boolean =>
+      !!kernelSpecs.specs &&
+      Object.keys(kernelSpecs.specs.kernelspecs).length > 0;
+
+    if (hasSpec()) {
+      return;
+    }
+
+    await new Promise<void>((resolve) => {
+      const onChange = (): void => {
+        if (hasSpec()) {
+          kernelSpecs.specsChanged.disconnect(onChange);
+          resolve();
+        }
+      };
+      kernelSpecs.specsChanged.connect(onChange);
+    });
+  }
+
+  /**
+   * Open the visible task notebook only after a kernel is available, so
+   * JupyterLab auto-selects the Pyodide kernel instead of prompting the user to
+   * pick one (CRT-399). The kernel-select dialog otherwise appears whenever the
+   * notebook opens before the Pyodide kernelspec has been registered.
+   */
+  private async openTaskNotebook(path: string): Promise<void> {
+    await this.waitForKernelSpecs();
+    this.documentManager.openOrReveal(path);
+  }
+
   async getTask(request: GetTask["request"]): Promise<Task> {
     try {
       // generate student task and autograder
@@ -204,7 +241,7 @@ export class EmbeddedPythonCallbacks {
       await this.setJupyterLocale(request.params.language);
 
       if (!isLoadTaskWithTask(request.params)) {
-        this.documentManager.openOrReveal(this.notebookToOpen);
+        await this.openTaskNotebook(this.notebookToOpen);
         return undefined;
       }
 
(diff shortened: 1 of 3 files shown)


apps/jupyter/extensions/notebook-runner/src/iframe-api.ts
```diff
@@ -125,6 +129,57 @@ export class EmbeddedPythonCallbacks {
...
+
+    await new Promise<void>((resolve) => {
+      function onChange(): void {
+        if (hasSpec()) {
+          clearTimeout(timeout);
+          kernelSpecs.specsChanged.disconnect(onChange);
+          resolve();
+        }
+      }
+
```
bubudetp, line 163: Given that we already have a way to know if the specs changed: 

https://github.com/jupyterlab/jupyterlab/blob/main/packages/services/src/kernelspec/manager.ts#L90

We don't necessarily need to reread the kernelspecs in the callback.

It should be enough if we do sth like:

`const hasSpec = (specs = kernelSpecs.specs): boolean =>
  !!specs?.kernelspecs?.[kernelName];`

and then

`const onSpecsChanged = (
  _sender: KernelSpec.IManager,
  specs: ISpecModels,
): void => {
  if (hasSpec(specs)) {
    cleanup();
    resolve();
  }
};`

because specsChanged emits the fetched specs object itself. It's also better to use the emitted argument instead of reading a mutable var and using it for resolving states

bubudetp (approved):
One non blocking comment, feel free to address it if you think it's reasonable

LGTM otherwise

apps/jupyter/extensions/notebook-runner/src/iframe-api.ts
pierluca: I fully agreed, and applied the comment, but the code became immediately more complex. I decided to revert in https://github.com/crt25/collimator/pull/615/changes/7a67a072d1555f866f01cc69fbfa4b93b0efce8a .
Let me know what you think.

apps/jupyter/extensions/notebook-runner/src/iframe-api.ts
```diff
@@ -166,6 +166,10 @@ export class EmbeddedPythonCallbacks {
       }, EmbeddedPythonCallbacks.kernelSpecWaitTimeoutMs);
 
       kernelSpecs.specsChanged.connect(onChange);
+
+      // re-check so that a spec registered between the check above and the
```
bubudetp: I'm not sure if this is possible, there is no await or any other async operation between the first check and specsChanged.connect and since javascript is single threaded, a race condition cannot happen here.

I'd personally drop that and keep where we read the specs from the signal.

apps/jupyter/extensions/notebook-runner/src/iframe-api.ts
```diff
@@ -125,6 +129,61 @@ export class EmbeddedPythonCallbacks {
       ? EmbeddedPythonCallbacks.taskTemplateLocation
       : EmbeddedPythonCallbacks.studentTaskLocation;
 
+  /**
+   * Resolves once at least one kernelspec (the Pyodide kernel) is registered,
```
bubudetp, line 133: Also the documentation here says it waits for the pyodide kernel but we resolve for any kernelspec in the code

apps/jupyter/extensions/notebook-runner/src/iframe-api.ts
```diff
@@ -125,6 +129,61 @@ export class EmbeddedPythonCallbacks {
...
+    await new Promise<void>((resolve) => {
+      function onChange(): void {
+        if (hasSpec()) {
+          clearTimeout(timeout);
+          kernelSpecs.specsChanged.disconnect(onChange);
+          resolve();
+        }
+      }
+
+      const timeout = setTimeout(() => {
```
bubudetp, line 164: should we clear the timeout here?

bubudetp (changes requested):

apps/jupyter/extensions/notebook-runner/src/iframe-api.ts
pierluca: I'm confused by this comment. Can you clarify what you have in mind?

apps/jupyter/extensions/notebook-runner/src/iframe-api.ts
pierluca: There's an await on our promise itself (which shouldn't cause execution to move to a different thread), and we call a bunch of external functions (including `kernelSpecs.specsChanged.connect`).

I agree that, as things stand, this may not be strictly necessary, but it seems like a reasonable defense given we call into external libraries.

bubudetp (approved):

apps/jupyter/extensions/notebook-runner/src/iframe-api.ts
bubudetp: I thought we had two different cleanup paths but I'm wrong, both behave identically

There's two completion paths now: 

The spec appears, onChange clears the pending timer, it disconnects and resolves the promise
The timout fires, the timer was already executing so it only needs to disconnect adn resolve

adding the timeout in the callback won't prevent anything, we only need to clear the timeout in the success path

PR #621, opened 2026-07-20
[CRT-363] Fix Jupyter teacher-notebook leak via mode-param collision

pierluca:
## Problem (CRT-363, security)

As a student in a Jupyter task, switching the UI language back and forth a few times replaces `task.ipynb` with the **teacher's notebook (solutions exposed)** and removes the autograder/student folders. [Repro video](https://jam.dev/c/24f627ba-dc72-4545-8633-a13c740013ba).

## Root cause — URL query-param namespace collision

Our extension reads the task mode from `?mode=`. But JupyterLite's `@jupyterlite/application-extension:mode-support` plugin **owns that same param** for the lab *shell* mode and rewrites it on every shell-mode change:

```js
labShell.modeChanged.connect((_, mode) => {
  url.searchParams.set("mode", mode);            // "solve" -> "single-document" -> "multiple-document"
  router.navigate(..., { skipRouting: true });
});
```

Our own `simplifyUserInterface` runs `application:toggle-mode` on every load, which fires `modeChanged` → `?mode=solve` is silently overwritten to `?mode=multiple-document`. Then **any plain reload of the iframe** (F5, "reload frame", history traversal — the extra history entries the rewrites push make Back reload at the rewritten URL) reads `multiple-document`, which `getModeFromUrl` doesn't recognize → falls through to the default `edit`. In edit mode, the frontend's re-sent `loadTask` (fired again on locale change) writes the **teacher template** into `/task.ipynb` and reveals it. The follow-up locale switch then navigates with `?mode=edit`, making it sticky — hence "3–5+ switches".

_(Full traced mechanism with file:line evidence and refuted alternatives is in the investigation notes; verified live that a load at `?mode=multiple-document` boots the extension in edit mode.)_

## Fix

1. **Rename our param `mode` → `crtMode`** — extension reader (`mode.ts`) + writer (`iframe-api.ts` locale navigation) + the six frontend Jupyter URL builders (`SolveTaskModal`, `EditTaskModal`, `ViewSolutionModal`, `CodeView`, `solve.tsx`, `view.tsx`). `mode-support` no longer touches our value, so it survives shell-mode toggles and reloads.
2. **Fail-safe default** — an absent/unknown `crtMode` now resolves to `show` (most restrictive), not `edit`. Edit must be requested explicitly, so no future param loss can ever expose the teacher template to a student. Every real flow sets `crtMode` explicitly, so teacher edit is unaffected.

Scratch is unaffected (routes by path, e.g. `/solve`, not a query param). Docs (`python.md`) updated.

## Verification

Frontend + extension `tsc --noEmit` and `eslint` clean; repo-wide sweep confirms no `mode=edit|solve|show` remains. **In-app smoke test still recommended** on a rebuilt JupyterLite: load `?crtMode=solve`, confirm it boots in solve mode and that `crtMode` survives the shell-mode toggle (unlike `mode`), and that repeated language switches no longer swap in the teacher notebook — I can run this next if you'd like.

---
## Summary by cubic
Fixes a security bug where language switches in Jupyter tasks could replace a student’s `task.ipynb` with the teacher notebook. Resolves CRT-363 by isolating our mode param, hardening parsing, and making the default safe.

- **Bug Fixes**
  - Switched URL param from `mode` to `crtMode` and centralized embedded-app URL building with `getEmbeddedAppUrl(...)` so Jupyter URLs consistently use `crtMode` and aren’t overwritten by `@jupyterlite/application-extension:mode-support` during shell-mode changes or reloads.
  - Default on missing/unknown `crtMode` is now `show`, and parsing uses an explicit switch (not enum indexing) to reject invalid values like `"0"` or `"constructor"`, preventing teacher notebook exposure.

<sup>Written for commit b936f720a226ae071dc0255a4637e25a1717cc18. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/crt25/collimator/pull/621?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/jupyter/extensions/notebook-runner/src/mode.ts
+++ b/apps/jupyter/extensions/notebook-runner/src/mode.ts
@@ -5,9 +5,18 @@ export enum Mode {
 }
 
 export const getModeFromUrl = (): Mode => {
-  // check if URL contains 'mode' parameter and retrieve the respective 'Mode' enum value, defaulting to 'edit'
+  // Read our own "crtMode" URL param. We must NOT use "mode": JupyterLite's
+  // `@jupyterlite/application-extension:mode-support` plugin owns that param and
+  // overwrites it with the lab shell mode ("single-document"/"multiple-document")
+  // whenever the shell mode changes — which our own `application:toggle-mode`
+  // call triggers on every load. Sharing the name let a reload read the shell
+  // mode, fall through to the default, and silently switch a student into edit
+  // mode, exposing the teacher notebook (CRT-363).
   const urlParams = new URLSearchParams(window.location.search);
-  const modeParam = urlParams.get("mode") ?? "edit";
+  const modeParam = urlParams.get("crtMode") ?? "show";
 
-  return Mode[modeParam as keyof typeof Mode] ?? Mode.edit;
+  // Default to the most restrictive mode. Edit must be requested explicitly, so
+  // a missing or unrecognized param can never expose the teacher template to a
+  // student — the frontend always sets crtMode explicitly for every flow.
+  return Mode[modeParam as keyof typeof Mode] ?? Mode.show;
 };
--- a/apps/jupyter/extensions/notebook-runner/src/mode.ts
+++ b/apps/jupyter/extensions/notebook-runner/src/mode.ts
@@ -13,10 +13,17 @@ export const getModeFromUrl = (): Mode => {
   // mode, fall through to the default, and silently switch a student into edit
   // mode, exposing the teacher notebook (CRT-363).
   const urlParams = new URLSearchParams(window.location.search);
-  const modeParam = urlParams.get("crtMode") ?? "show";
 
   // Default to the most restrictive mode. Edit must be requested explicitly, so
   // a missing or unrecognized param can never expose the teacher template to a
   // student — the frontend always sets crtMode explicitly for every flow.
-  return Mode[modeParam as keyof typeof Mode] ?? Mode.show;
+  switch (urlParams.get("crtMode")) {
+    case "edit":
+      return Mode.edit;
+    case "solve":
+      return Mode.solve;
+    case "show":
+    default:
+      return Mode.show;
+  }
 };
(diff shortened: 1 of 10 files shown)


apps/jupyter/extensions/notebook-runner/src/mode.ts
```diff
@@ -5,9 +5,18 @@ export enum Mode {
...
+  // mode, exposing the teacher notebook (CRT-363).
   const urlParams = new URLSearchParams(window.location.search);
-  const modeParam = urlParams.get("mode") ?? "edit";
+  const modeParam = urlParams.get("crtMode") ?? "show";
 
-  return Mode[modeParam as keyof typeof Mode] ?? Mode.edit;
+  // Default to the most restrictive mode. Edit must be requested explicitly, so
+  // a missing or unrecognized param can never expose the teacher template to a
+  // student — the frontend always sets crtMode explicitly for every flow.
+  return Mode[modeParam as keyof typeof Mode] ?? Mode.show;
```
bubudetp: This is not fail-safe Mode[modeParam]

for an 

enum Mode {
edit,
solve, 
show,
}

js creates {edit: 0, solve: 1, show: 2, 0: "edit", 1: "solve", 2: "show"}

`Mode[modeParam as keyof typeof Mode] ?? Mode.show` won't validate that the result is a valid mode, it will only fall back when the lookup returns null or undefined

we could replace this with a switch

```suggestion
  switch (urlParams.get("crtMode")) {
    case "edit":
      return Mode.edit;
    case "solve":
      return Mode.solve;
    case "show":
    default:
      return Mode.show;
  }
```

frontend/src/components/dashboard/CodeView.tsx
```diff
@@ -41,7 +41,7 @@ const getSolutionCodeUrl = (taskType: TaskType) => {
     case TaskType.SCRATCH:
       return `${scratchAppHostName}/show`;
     case TaskType.JUPYTER:
-      return `${jupyterAppHostName}?mode=show`;
+      return `${jupyterAppHostName}?crtMode=show`;
```
bubudetp: This is duplicated 6 times, having a `getEmbeddedAppUrl(taskType, mode)` utility would centralise things

bubudetp (changes requested):

bubudetp (approved):

PR #622, opened 2026-07-21
[CRT-397] Fix Jupyter locale change losing work + student-path content error

pierluca:
## Problem (CRT-397)

In the student Jupyter interface, making unsaved changes and then **changing the UI language** (a) loses the unsaved work and (b) throws *"could not find content with path student/task.ipynb"* when the student clicks the task in the file tree. [Jam](https://jam.dev/c/7eb26ff7-8396-4758-bfa8-315b24b5f2c4).

## Root cause

Changing the language reloads the iframe (JupyterLab applies locale bundles only at load). JupyterLite contents are **memory-backed**, so the reload wipes the workspace. Two independent defects surface:

**1. The content-not-found error.** In solve mode the visible root `task.ipynb` is the student's *handle* to their task: `redirectTaskFileOpensToStudentVersion` transparently redirects a click on it to `/student/task.ipynb`. After the reload wipes memory, `/student/task.ipynb` hasn't been re-written yet (the frontend re-sends it asynchronously), so the redirect opens a path that doesn't exist → the error.

**2. Lost unsaved changes.** The pre-reload auto-save pushes the current solution up (`postTaskSolution`), but `EmbeddedApp`'s handler doesn't await the backend write, so the iframe navigates while the POST is in flight. After the reload, `onAppAvailable`'s `fetchLatestSolutionFile` races that POST and reads the **stale** solution.

## Fix

- **`user-interface.ts`** — **guard** the student-task redirect: open `/student/task.ipynb` only once it exists (an async `contents.get` check); otherwise do nothing (the task auto-opens as soon as the content is restored). `task.ipynb` stays visible — it's the student's own entry point — and the template is never opened.
- **`solve.tsx`** — stash the solution the embedded app pushes (`onReceiveTaskSolution`) in a `useRef`. The parent frame isn't reloaded, so the ref survives; `onAppAvailable` **replays the stashed solution** after the reload instead of re-fetching, so the freshest edits survive.

No change to the reload mechanism, so this is independent of CRT-363's `mode`→`crtMode` rename.

## Verification

Frontend + extension `tsc --noEmit` and `eslint` clean. **In-app smoke test recommended** on a rebuilt JupyterLite: as a student, edit → change language → confirm edits survive, no content error, and `task.ipynb` still opens the student copy.

### DIFF
--- a/apps/jupyter/extensions/notebook-runner/src/user-interface.ts
+++ b/apps/jupyter/extensions/notebook-runner/src/user-interface.ts
@@ -170,15 +170,11 @@ const isTaskFile = (path: string): boolean =>
   path === "/task.ipynb" || path === "task.ipynb";
 
 /**
- * The student never edits the teacher template directly: opening `task.ipynb`
- * transparently opens their own `/student/task.ipynb`. That copy is (re)written
- * asynchronously after a reload (e.g. a language change wipes the memory-backed
- * contents and the frontend re-sends the submission), so opening it too early
- * throws "could not find content" (CRT-397). We therefore open it only if it
- * already exists; if it does not, the open is DROPPED — there is no waiting or
- * retrying here. That is safe because the restore flow (loadSubmission) issues
- * its own open after writing the file, and we must never fall back to opening
- * the template itself.
+ * Entry point to open the student notebook, only if it's available.
+ * In case of failure, we let the restore flow (loadSubmission) issue
+ * its own open after writing the file.
+ * Without this, quick locale change could lead to failure when
+ * opening the notebook, or worse, opening the wrong task.
  */
 const redirectTaskFileOpensToStudentVersion = (
   documentManager: IDocumentManager,
@@ -199,9 +195,6 @@ const redirectTaskFileOpensToStudentVersion = (
       .then(
         () => open(studentTaskPath, widgetName, kernel, options),
         (error) => {
-          // Only contents.get() rejections land here: the student copy has not
-          // been restored yet, so this open is dropped (the restore flow opens
-          // the task itself once the content is written).
           console.debug(
             `Ignoring open of ${studentTaskPath}; not available yet`,
             error,
--- a/apps/jupyter/extensions/notebook-runner/src/user-interface.ts
+++ b/apps/jupyter/extensions/notebook-runner/src/user-interface.ts
@@ -195,8 +195,16 @@ const redirectTaskFileOpensToStudentVersion = (
     void documentManager.services.contents
       .get(studentTaskPath, { content: false })
       .then(() => open(studentTaskPath, widgetName, kernel, options))
-      .catch(() => {
-        // the student copy has not been restored yet; ignore this open.
+      .catch((error) => {
+        // The expected case: the student copy has not been restored yet, so we
+        // ignore this open (the task auto-opens once the content is restored).
+        // Contents are memory-backed in JupyterLite so other failures are not
+        // really expected; log a breadcrumb instead of staying fully silent in
+        // case one ever occurs.
+        console.debug(
+          `Ignoring open of ${studentTaskPath}; not available yet`,
+          error,
+        );
       });
   };
 
--- a/apps/jupyter/extensions/notebook-runner/src/user-interface.ts
+++ b/apps/jupyter/extensions/notebook-runner/src/user-interface.ts
@@ -174,14 +174,16 @@ const isTaskFile = (path: string): boolean =>
  * transparently opens their own `/student/task.ipynb`. That copy is (re)written
  * asynchronously after a reload (e.g. a language change wipes the memory-backed
  * contents and the frontend re-sends the submission), so opening it too early
- * throws "could not find content" (CRT-397). We therefore only open it once it
- * exists, and otherwise do nothing — the task auto-opens as soon as the content
- * is restored, and we must never fall back to opening the template itself.
+ * throws "could not find content" (CRT-397). We therefore open it only if it
+ * already exists; if it does not, the open is DROPPED — there is no waiting or
+ * retrying here. That is safe because the restore flow (loadSubmission) issues
+ * its own open after writing the file, and we must never fall back to opening
+ * the template itself.
  */
 const redirectTaskFileOpensToStudentVersion = (
   documentManager: IDocumentManager,
 ): void => {
-  const openStudentTaskWhenReady = (
+  const openStudentTaskIfAvailable = (
     open: (
       path: string,
       widgetName?: string,
@@ -194,17 +196,21 @@ const redirectTaskFileOpensToStudentVersion = (
   ): void => {
     void documentManager.services.contents
       .get(studentTaskPath, { content: false })
-      .then(() => open(studentTaskPath, widgetName, kernel, options))
+      .then(
+        () => open(studentTaskPath, widgetName, kernel, options),
+        (error) => {
+          // Only contents.get() rejections land here: the student copy has not
+          // been restored yet, so this open is dropped (the restore flow opens
+          // the task itself once the content is written).
+          console.debug(
+            `Ignoring open of ${studentTaskPath}; not available yet`,
+            error,
+          );
+        },
+      )
       .catch((error) => {
-        // The expected case: the student copy has not been restored yet, so we
-        // ignore this open (the task auto-opens once the content is restored).
-        // Contents are memory-backed in JupyterLite so other failures are not
-        // really expected; log a breadcrumb instead of staying fully silent in
-        // case one ever occurs.
-        console.debug(
-          `Ignoring open of ${studentTaskPath}; not available yet`,
-          error,
-        );
+        // reached only if opening the existing file itself failed
+        console.error(`Failed to open ${studentTaskPath}`, error);
       });
   };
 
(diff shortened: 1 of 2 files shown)


frontend/src/pages/class/[classId]/session/[sessionId]/task/[taskId]/solve.tsx
```diff
@@ -257,6 +301,11 @@ const SolveTaskPage = () => {
         return;
       }
 
+      // Stash synchronously so a reload can replay it even if the backend write
```
Comment (bubudetp) [inline frontend/src/pages/class/[classId]/session/[sessionId]/task/[taskId]/solve.tsx:287]: currently, the student work is stored in pendingSolution, the same work is saved to the backend but pendingSolution is not cleared. 

it should because later, if the iframe reloads, the page reuses pendingSolution without checking the backend..

frontend/src/pages/class/[classId]/session/[sessionId]/task/[taskId]/solve.tsx
```diff
@@ -237,6 +272,13 @@ const SolveTaskPage = () => {
           { intl, descriptor: taskMessages.cannotLoadSubmission },
         );
       } catch {
+        // Put the stash back (unless something newer arrived meanwhile) so the
+        // next app load can still restore it — it may hold work that never
+        // reached the backend. A stash for a different task is dropped instead.
+        if (stashedSolution !== null) {
```
Comment (bubudetp) [inline frontend/src/pages/class/[classId]/session/[sessionId]/task/[taskId]/solve.tsx:260]: if the stash entry exists but loading fails, we shouldn't call load task, we should keep the notebook disabled or have a recovery screen and retry automatically, if the retry fails, indicate some recoverage message on the UI.

because a failure to load the stashed copy does not mean the copy is bad or throwable. In this context (after a locale change), the iframe might be still reloading, jupyter still might be initializing..

apps/jupyter/extensions/notebook-runner/src/user-interface.ts
```diff
@@ -155,13 +155,49 @@ const hideGradingFolders = (fileBrowser: FileBrowser): void => {
...
+ * is restored, and we must never fall back to opening the template itself.
+ */
 const redirectTaskFileOpensToStudentVersion = (
   documentManager: IDocumentManager,
 ): void => {
-  const redirect = (path: string): string =>
-    path === "/task.ipynb" || path === "task.ipynb"
-      ? "/student/task.ipynb"
-      : path;
+  const openStudentTaskWhenReady = (
```
bubudetp: it sounds like this function keeps waiting and opens the task once the file appears but it doesn't, it checks if the file exists right now, if yes it opens it, if not, it logs it and stops

apps/jupyter/extensions/notebook-runner/src/user-interface.ts
```diff
@@ -155,13 +155,49 @@ const hideGradingFolders = (fileBrowser: FileBrowser): void => {
...
+      path: string,
+      widgetName?: string,
+      kernel?: Partial<Kernel.IModel>,
+      options?: DocumentRegistry.IOpenOptions,
+    ) => IDocumentWidget | undefined,
+    widgetName?: string,
+    kernel?: Partial<Kernel.IModel>,
+    options?: DocumentRegistry.IOpenOptions,
+  ): void => {
+    void documentManager.services.contents
```
bubudetp, line 193: it's a bit tricky here but you're using the same catch block for contents.get() and open() which means there might be many cases that can be logged as "not available yet" such as the file not existing, or perhaps jupyter's content service not being available, or the file existing but open() failing and much more.

bubudetp (changes requested):

frontend/src/pages/class/[classId]/session/[sessionId]/task/[taskId]/solve.tsx
Comment (pierluca) [reply frontend/src/pages/class/[classId]/session/[sessionId]/task/[taskId]/solve.tsx:287]: I'm confused, isn't this exactly what we want to solve this problem?

frontend/src/pages/class/[classId]/session/[sessionId]/task/[taskId]/solve.tsx
Comment (pierluca) [reply frontend/src/pages/class/[classId]/session/[sessionId]/task/[taskId]/solve.tsx:260]: Can't believe I missed this when reviewing. Good point!

apps/jupyter/extensions/notebook-runner/src/user-interface.ts
pierluca: does it matter here? 🤔 not sure I see it - it's in any case an open that won't go through because something's unavailable, and we're logging the error.

apps/jupyter/extensions/notebook-runner/src/user-interface.ts
pierluca: You're right, I was working under the assumption that `documentManager.services.contents.get()` returned a promise and would guarantee either a definitive failure or a "wait until this works".
It's not the semantic, see the implementation here
https://github.com/jupyterlab/jupyterlab/blob/81d1026dcd96b75fb4f80e066ecb51d7428fce7e/packages/services/src/contents/index.ts#L1263

pierluca (commented):
Thanks for your in-depth review!

apps/jupyter/extensions/notebook-runner/src/user-interface.ts
```diff
@@ -164,13 +164,55 @@ const hideGradingFolders = (fileBrowser: FileBrowser): void => {
...
+/**
+ * The student never edits the teacher template directly: opening `task.ipynb`
+ * transparently opens their own `/student/task.ipynb`. That copy is (re)written
+ * asynchronously after a reload (e.g. a language change wipes the memory-backed
+ * contents and the frontend re-sends the submission), so opening it too early
+ * throws "could not find content" (CRT-397). We therefore open it only if it
+ * already exists; if it does not, the open is DROPPED — there is no waiting or
+ * retrying here. That is safe because the restore flow (loadSubmission) issues
+ * its own open after writing the file, and we must never fall back to opening
+ * the template itself.
```
pierluca:
```suggestion
 * Entry point to open the student notebook, only if it's available.
 * In case of failure, we let the restore flow (loadSubmission) issue
 * its own open after writing the file.
 * Without this, quick locale change could lead to failure when
 * opening the notebook, or worse, opening the wrong task.
```

apps/jupyter/extensions/notebook-runner/src/user-interface.ts
```diff
@@ -164,13 +164,55 @@ const hideGradingFolders = (fileBrowser: FileBrowser): void => {
...
+  ): void => {
+    void documentManager.services.contents
+      .get(studentTaskPath, { content: false })
+      .then(
+        () => open(studentTaskPath, widgetName, kernel, options),
+        (error) => {
+          // Only contents.get() rejections land here: the student copy has not
+          // been restored yet, so this open is dropped (the restore flow opens
+          // the task itself once the content is written).
+          console.debug(
```
pierluca:
```suggestion
          console.debug(
```

frontend/src/pages/class/[classId]/session/[sessionId]/task/[taskId]/solve.tsx
```diff
@@ -96,6 +96,27 @@ const SolveTaskPage = () => {
...
   const wasInitialized = useRef(false);
   const isScratchMutexAvailable = useRef(true);
+  // The freshest solution the embedded app has pushed up (e.g. the auto-save
+  // triggered right before a language change reloads the iframe). The parent
+  // frame is not reloaded, so this ref survives; we replay it after the reload
+  // instead of racing the still-in-flight backend write with a stale fetch,
+  // which otherwise loses the student's unsaved changes (CRT-397). Keyed by
+  // task id: this page component survives task-to-task navigation (dynamic
+  // route params do not remount it), so an unconsumed stash must never be
+  // replayed into a different task.
```
Comment (pierluca) [inline frontend/src/pages/class/[classId]/session/[sessionId]/task/[taskId]/solve.tsx]:
```suggestion
  // The freshest solution the embedded app has pushed up (e.g. the auto-save
  // triggered right before a language change reloads the iframe).
  // This allows data persistency after app iframe reloads (e.g. locale change).
  // Keyed by task id: this page component survives task-to-task navigation,
  // so an unconsumed stash must never be replayed into a different task.
```

frontend/src/pages/class/[classId]/session/[sessionId]/task/[taskId]/solve.tsx
```diff
@@ -225,6 +260,18 @@ const SolveTaskPage = () => {
...
         );
       } catch {
+        // Put the stash back (unless something newer arrived meanwhile) so the
+        // next app load can still restore it — it may hold work that never
+        // reached the backend. A stash for a different task is dropped instead.
+        // A failed replay does not mean the stashed copy is bad (the embedded
+        // app may still be initializing after a locale change), so we must NOT
+        // fall back to the pristine task in that case — it would visibly
+        // discard the stashed work; the next app load replays the stash.
+        if (stashedSolution !== null) {
```
Comment (pierluca) [inline frontend/src/pages/class/[classId]/session/[sessionId]/task/[taskId]/solve.tsx]:
```suggestion
        if (stashedSolution !== null) {
```

frontend/src/pages/class/[classId]/session/[sessionId]/task/[taskId]/solve.tsx
```diff
@@ -245,6 +294,11 @@ const SolveTaskPage = () => {
         return;
       }
 
+      // Stash synchronously so a reload can replay it even if the backend write
+      // below is still in flight (CRT-397). Keyed by task id so it can never be
+      // replayed into a different task.
```
Comment (pierluca) [inline frontend/src/pages/class/[classId]/session/[sessionId]/task/[taskId]/solve.tsx]:
```suggestion
      // Stash synchronously so a reload can replay it even if the backend write
      // below is still in flight.
```

bubudetp (approved):

PR #626, opened 2026-07-27
[CRT-442] Scope student writes to the sessions they take part in

pierluca:
POST /classes/:classId/sessions/:sessionId/task/:taskId/solutions/student and POST /student-activity performed no authorization check at all. `@StudentOnly()` only proves that *some* student is authenticated; both handlers then took the session and task straight from the request (the URL for solutions, the body for activities) and the student from the token. The solutions route even discarded its classId parameter, so any student token could persist a solution into any lesson of any class - including a different teacher's - as long as the lesson contained the task. Every read path in these controllers calls authorizationService.canView...; these two write paths called nothing.

Add AuthorizationService.canCreateStudentSolution and canTrackStudentActivities, both built on isStudentOfSessionTask: the task must be part of the lesson (a live SessionTask row) and the student must take part in that lesson, either by having joined it anonymously or by being an authenticated student of the lesson's class. For solutions the lesson must additionally belong to the classId in the route, which is no longer discarded. Deny returns 403.

⚠️ The check deliberately ignores Task.deletedAt: the SessionTask row is what makes a task part of a lesson, and a student may legitimately still be working on a task that was soft-deleted after the lesson started.

❓ Desirable? It doesn't seem wrong to me.

---
## Summary by cubic
Scopes student write endpoints to only the sessions they participate in to fix CRT-442. Submitting solutions or tracking activities outside the student’s session now returns 403.

- **Bug Fixes**
  - Added `AuthorizationService.canCreateStudentSolution` and `canTrackStudentActivities`, built on a shared `isStudentOfSessionTasks` helper; participation via anonymous join or class membership; ignores `Task.deletedAt`.
  - `POST /classes/:classId/sessions/:sessionId/task/:taskId/solutions/student` now enforces `classId`/`sessionId`/`taskId` and class scoping; returns 403 if not allowed.
  - `POST /student-activity` batch-validates distinct session/task pairs in one query; rejects the whole batch if any pair is out of scope; allows empty batches without DB calls. Activities are not scoped to a class; solutions are.
  - Strengthened tests to cover single-query lookups, duplicate-pair de-duplication, soft-delete filters, and unauthenticated short-circuit.

<sup>Written for commit 5e493549002d843beb7ff605e1368a3311261b37. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/crt25/collimator/pull/626?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/src/api/authorization/authorization.service.ts
+++ b/backend/src/api/authorization/authorization.service.ts
@@ -310,6 +310,38 @@ export class AuthorizationService {
     return sessionTask !== null;
   }
 
+  protected async isStudentOfSessionTasks(
+    authenticatedStudent: Student,
+    sessionTasks: Array<{ sessionId: number; taskId: number }>,
+    classId?: number,
+  ): Promise<boolean> {
+    if (sessionTasks.length === 0) {
+      return true;
+    }
+
+    const uniquePairs = [
+      ...new Map(
+        sessionTasks.map((pair) => [`${pair.sessionId}/${pair.taskId}`, pair]),
+      ).values(),
+    ];
+
+    const authorizedPairs = await this.prisma.sessionTask.findMany({
+      select: { sessionId: true, taskId: true },
+      where: {
+        OR: uniquePairs,
+        deletedAt: null,
+        session: {
+          ...(classId === undefined ? {} : { classId }),
+          deletedAt: null,
+          class: { deletedAt: null },
+          ...this.participatesInSession(authenticatedStudent.id),
+        },
+      },
+    });
+
+    return authorizedPairs.length === uniquePairs.length;
+  }
+
   async canCreateStudentSolution(
     authenticatedStudent: Student | null,
     classId: number,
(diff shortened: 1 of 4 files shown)


backend/src/api/authorization/authorization.service.ts
```diff
@@ -255,6 +255,120 @@ export class AuthorizationService {
...
+    );
+
+    const results = await Promise.all(
+      [...distinctSessionTasks.values()].map(({ sessionId, taskId }) =>
+        this.isStudentOfSessionTask(authenticatedStudent, sessionId, taskId),
+      ),
+    );
+
+    const canTrack = results.every((isAuthorized) => isAuthorized);
+
```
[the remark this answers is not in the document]
bubudetp: Good point but bad suggestion

bubudetp (approved):
Fix LGTM

I took this opportunity to optimize the student authorization lookup and improve the test coverage by asserting exact expected queries to be called and more..

backend/src/api/authorization/authorization.service.ts
```diff
@@ -255,6 +255,151 @@ export class AuthorizationService {
...
+          ...(classId === undefined ? {} : { classId }),
+          deletedAt: null,
+          class: { deletedAt: null },
+          ...this.participatesInSession(authenticatedStudent.id),
+        },
+      },
+    });
+
+    return sessionTask !== null;
+  }
```
pierluca: do we still need the single-task version of this function, considering we now have `isStudentOfSessionTasks` ? If not, I would remove the single-task version.

Furthermore, given these are not entirely trivial, I would also consider copying over the comments from `isStudentOfSessionTask` to `isStudentOfSessionTasks`, as the comments are rather useful.

bubudetp: Nothing prevents us from reusing the `isStudentOfSessionsTasks`. Fair point on the comments 👌🏼

PR #627, opened 2026-07-27
[CRT-443] Do not let students read a task's reference solutions

pierluca:
GET /tasks/:id/with-reference-solutions returned each reference solution's file (base64) and its tests - the task's answer key - to anyone allowed on the endpoint, and the endpoint allowed STUDENT. Any authenticated student could read the answers to any task by id.

Remove STUDENT from the endpoint's @Roles. Only the teacher-facing task detail and reference-solution pages use it; the student flow reads the task through :id and :id/download, which are unaffected.

Enforcement is the @Roles decorator read by the global RoleGuard, so the test drives the guard against the real controller metadata: a student is denied on :id/with-reference-solutions (ForbiddenException) while still allowed on the plain :id, and teacher/admin remain allowed. Confirmed the deny case fails without this change (the student resolves instead of being rejected) and passes with it.

---
## Summary by cubic
Block students from reading reference solutions by restricting `GET /tasks/:id/with-reference-solutions` to `ADMIN` and `TEACHER` only. Addresses CRT-443.

- **Bug Fixes**
  - Scopes allowed roles exactly to `[ADMIN, TEACHER]`; removes `STUDENT`. Student `GET /tasks/:id` and `:id/download` remain allowed.
  - Adds authorization tests using real controller metadata and production-like `isStudent`; students are forbidden, teachers/admins allowed.

<sup>Written for commit c6935c4b7703327adcfb3470415f312706218b94. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/crt25/collimator/pull/627?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/src/api/tasks/reference-solutions-authorization.spec.ts
+++ b/backend/src/api/tasks/reference-solutions-authorization.spec.ts
@@ -55,6 +55,18 @@ describe("TasksController reference-solutions authorization", () => {
     TasksController.prototype.findOneWithReferenceSolutions;
   const taskHandler = TasksController.prototype.findOne;
 
+  it("scopes reference solutions to exactly teachers and admins", () => {
+    const allowedRoles = reflector.get<Role[]>(
+      ALLOWED_ROLES,
+      referenceSolutionsHandler,
+    );
+
+    expect(allowedRoles).toHaveLength(2);
+    expect(allowedRoles).toEqual(
+      expect.arrayContaining([UserType.ADMIN, UserType.TEACHER]),
+    );
+  });
+
   it("denies a student reading a task's reference solutions", async () => {
     const guard = buildGuard(student, true);
 
--- a/backend/src/api/tasks/reference-solutions-authorization.spec.ts
+++ b/backend/src/api/tasks/reference-solutions-authorization.spec.ts
@@ -1,0 +1,91 @@
+import { ExecutionContext, ForbiddenException } from "@nestjs/common";
+import { Reflector } from "@nestjs/core";
+import { Student, User, UserType } from "@prisma/client";
+import { RoleGuard } from "../authentication/role.guard";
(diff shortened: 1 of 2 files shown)


backend/src/api/tasks/reference-solutions-authorization.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+// The reference-solutions endpoint returns each reference solution's file and
+// tests - the task's answer key - so it must be teacher/admin only. Enforcement
+// is entirely the @Roles decorator read by the global RoleGuard, so this
+// exercises the guard against the real controller metadata: a student is denied
+// on GET :id/with-reference-solutions but still allowed on the plain GET :id
+// they legitimately use.
+describe("TasksController reference-solutions authorization", () => {
+  const buildGuard = (
+    principal: User | Student,
+    isStudent: boolean,
```
bubudetp, line 17: This is weird, in theory this says you can be a teacher, but also a student where you can set -> principal: User and isStudent = True

backend/src/api/tasks/reference-solutions-authorization.spec.ts
```diff
@@ -0,0 +1,91 @@
+import { ExecutionContext, ForbiddenException } from "@nestjs/common";
```
bubudetp, line 1: there's no test which asserts the exact role scope (Admin and Teacher, nothing else)

bubudetp (approved):

PR #629, opened 2026-07-27
[CRT-445] Refuse editing a public task other users' lessons depend on

pierluca:
TasksService.deleteById refuses to delete a public task that another user's lesson uses (TASK_IN_OTHER_USERS_LESSON); TasksService.update had no equivalent guard, so the same task could be edited out from under that lesson - the other teacher's lesson silently starts serving a different exercise, and update also drops the reference solutions no longer listed. Apply the same guard on the update path and map it to 409 in the controller as delete already does; own lessons are unaffected. The frontend needs no change - it already has a translated message for the error code.

---
## Summary by cubic
Prevents editing a public task if another teacher’s lesson depends on it, matching the delete rule. Returns 409 with TASK_IN_OTHER_USERS_LESSON; private tasks and own lessons are unaffected. Addresses CRT-445.

- **Bug Fixes**
  - Added guard in `TasksService.update` to throw `TaskInOtherUsersLessonError` for public tasks used by other users’ lessons.
  - Mapped the error to 409 Conflict in `TasksController.update` (`TASK_IN_OTHER_USERS_LESSON`).
  - Fixed a regression where a local `task` variable shadowed the update input; renamed to `existingTask` so field changes persist.
  - Added tests for service behavior (public vs private, check ordering, field persistence) and controller conflict mapping.

<sup>Written for commit 65c04856fcf5d50c7bd30337354a085c8ef1768d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/crt25/collimator/pull/629?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/src/api/tasks/task-update.spec.ts
+++ b/backend/src/api/tasks/task-update.spec.ts
@@ -15,6 +15,32 @@ import {
 // lesson with students" checks (called in that order for a public task).
 describe("TasksService.update", () => {
   const taskId = 1;
+  const taskData = new Uint8Array([1, 2, 3]);
+  const updatedTask = { id: taskId };
+  const studentUsageQuery = {
+    where: {
+      taskId,
+      task: { deletedAt: null },
+      session: {
+        deletedAt: null,
+        OR: [
+          {
+            anonymousStudents: {
+              some: { deletedAt: null },
+            },
+          },
+          {
+            class: {
+              deletedAt: null,
+              students: {
+                some: { deletedAt: null },
+              },
+            },
+          },
+        ],
+      },
+    },
+  };
 
   const buildService = (opts: {
     isPublic: boolean;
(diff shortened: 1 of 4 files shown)


backend/src/api/tasks/task-update.spec.ts
```diff
@@ -0,0 +1,141 @@
+import { PrismaService } from "src/prisma/prisma.service";
```
bubudetp, line 1: While we're adding tests, we should add one which covers the controller's new 409 branch

backend/src/api/tasks/task-update.spec.ts
```diff
@@ -0,0 +1,141 @@
...
+    expect(tx.task.update).toHaveBeenCalledWith(
+      expect.objectContaining({
+        data: expect.objectContaining({
+          title: "rewritten",
+          description: "new",
+        }),
+      }),
+    );
+  });
+});
```
bubudetp: This assertion is too broad, we're only checking two fields over the following:

where
mimeType
task file data
omit
reference solutions
...

bubudetp (approved):

PR #644, opened 2026-07-29
[CRT-459] Fix four Python-converter defects on valid student code

pierluca:
Supersedes #641 (auto-closed during the branch rename to the ticket name — same content, commits now labeled [CRT-459]).

## Summary

Follow-up to #639: an adversarial review of the whole Python AST conversion (grammar-vs-code analysis of every ctx accessor by three independent reviewers, cross-checked by a 106-construct valid-Python corpus run through the converter) found four more defects, all demonstrated by failing tests before being fixed.

### Crashes on valid Python (`TypeError: Cannot read properties of null (reading 'accept')`)

1. **Bare `yield`** — the guard added in `521a772a` runs *after* `getExpression(null)` has already thrown, so it was unreachable and a bare `yield` still crashed. The null-check now happens before.
2. **`a[:]` / `a[::]`** — a zero-expression slice fell through to the `named_expression` branch, which is null for the colon form.
3. **`f(x=2, *args)`** — star-unpacking after a keyword argument takes the `kwarg_or_starred: starred_expression` alternative, whose `expression()` accessor is null yet was read unconditionally (its hand-built `star-args` branch was unreachable dead code).

### Silent wrong ASTs (no crash — the similarity analysis was just wrong)

4. **`a[::2]` (step) ≡ `a[:2]` (stop)** — slice expressions were bucketed by count, not by which colon they follow. Now bucketed by colon position.
5. **`def` ≡ `async def`** — `isAsync` was computed with `ctx.ASYNC() !== undefined` (always true; ANTLR returns null) and then never surfaced into the AST at all, while `async for`/`async with` *are* distinguished. Fixed the detection and added an optional `isAsync` marker to `FunctionDeclarationNode`, set only for async declarations (existing sync ASTs unchanged).

## Commits

- `[CRT-459]` failing tests for all five behaviours + a permanent **crash corpus**: 106 valid Python constructs that must all convert without throwing.
- Four `[CRT-459]` fix commits, one per defect area (slice, kwarg-or-starred, yield, async).

## Verification

- All new tests fail before their fix and pass after.
- `src/ast`: 15 suites, 280 tests, green. Full backend jest: only pre-existing DB-dependent suites fail locally (`PrismaClientInitializationError`, no test DB in this environment) — unrelated to this `src/ast`-only diff.
- eslint + prettier clean.

### Out of scope (identified, fixed separately)

Uncontrolled behaviour on **invalid** Python — see [CRT-458] #643.

[CRT-459]: https://classroomrt.atlassian.net/browse/CRT-459?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CRT-459]: https://classroomrt.atlassian.net/browse/CRT-459?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---
## Summary by cubic
Fixes four defects in the Python AST converter that crashed or produced wrong ASTs on valid code, and adds a 106-case valid-Python crash corpus to prevent regressions. Addresses CRT-459.

- **Bug Fixes**
  - Bare `yield` no longer crashes; converts to a `yield` operator with no operands.
  - Slice parsing now handles `a[:]` and `a[::]`, and buckets by colon position so `a[:2]` (stop) and `a[::2]` (step) produce different ASTs; tests include nested bounds to count only the slice’s own colons.
  - `f(x=2, *args)` now visits `starred_expression`, avoiding a null read and producing the same unpack operator as `f(*args)`.

<sup>Written for commit 51db8878db7fbb5379b3b76ed3a6a28a5669dc49. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/crt25/collimator/pull/644?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/src/ast/converters/python/nodes/expressions/kwarg-or-starred.ts
+++ b/backend/src/ast/converters/python/nodes/expressions/kwarg-or-starred.ts
@@ -1,14 +1,8 @@
-import { AstNodeType } from "src/ast/types/general-ast";
-import {
-  ExpressionNodeType,
-  OperatorNode,
-} from "src/ast/types/general-ast/ast-nodes/expression-node";
 import { IPythonAstVisitor } from "../../python-ast-visitor-interface";
 import {
   Kwarg_or_starredContext,
   NameContext,
 } from "../../generated/PythonParser";
-import { starArgsOperator } from "../../operators";
 import { PythonFunctionArgument } from "./args";
 
 export const convertKwargOrStarred = (
@@ -16,22 +10,20 @@ export const convertKwargOrStarred = (
   ctx: Kwarg_or_starredContext,
 ): PythonFunctionArgument => {
   const name = ctx.name() as NameContext | undefined;
-  const expression = visitor.getExpression(ctx.expression());
 
   if (name) {
     return {
       name: name.getText(),
-      expression: expression.node,
+      expression: visitor.getExpression(ctx.expression()).node,
     };
   }
 
+  // The grammar's other alternative: a starred_expression (`*args` appearing
+  // after a keyword argument has started the kwargs list). expression() is
+  // null here; visiting the starred expression produces the same unpack
+  // operator as a plain `f(*args)` call.
   return {
     name: null,
-    expression: {
-      nodeType: AstNodeType.expression,
-      expressionType: ExpressionNodeType.operator,
-      operator: starArgsOperator,
-      operands: [expression.node],
-    } satisfies OperatorNode,
+    expression: visitor.getExpression(ctx.starred_expression()).node,
   };
 };
--- a/backend/src/ast/converters/python/nodes/expressions/slice.ts
+++ b/backend/src/ast/converters/python/nodes/expressions/slice.ts
@@ -13,48 +13,45 @@ export const convertSlice = (
   visitor: IPythonAstVisitor,
   ctx: SliceContext,
 ): PythonVisitorReturnValue => {
-  const { nodes: expressions, functionDeclarations } = visitor.getExpressions(
-    ctx.expression_list(),
-  );
+  const colons = ctx.COLON_list();
 
-  if (expressions.length > 0) {
-    let startExpression: ExpressionNode | null = null;
-    let stopExpression: ExpressionNode | null = null;
-    let stepExpression: ExpressionNode | null = null;
-
-    const colonList = ctx.COLON_list();
-    if (ctx.start.text === ":") {
-      // we do not have a start expression
-      stopExpression = expressions[0];
-      stepExpression = expressions.length > 1 ? expressions[1] : null;
-    } else if (colonList.length === 1) {
-      // we do not have a step expression
-      startExpression = expressions[0];
-      stopExpression = expressions.length > 1 ? expressions[1] : null;
-    } else if (colonList.length === 2 && expressions.length === 2) {
-      // we have do not have a stop expression
-      startExpression = expressions[0];
-      stepExpression = expressions[1];
-    } else if (colonList.length === 2 && expressions.length === 3) {
-      // we have all three expressions
-      startExpression = expressions[0];
-      stopExpression = expressions[1];
-      stepExpression = expressions[2];
-    } else {
-      throw new Error(
-        `Unexpected slice format. Found ${colonList.length} colons, ${expressions.length} expressions and the '${ctx.start.text}' starting token.`,
-      );
-    }
-
-    return createSliceExpression(
-      startExpression,
-      stopExpression,
-      stepExpression,
-      functionDeclarations,
-    );
+  if (colons.length === 0) {
+    // the grammar's named_expression alternative: a plain subscript like a[1]
+    return visitor.visit(ctx.named_expression());
   }
 
-  return visitor.visit(ctx.named_expression());
+  // The colon form: which of start/stop/step an expression is follows from
+  // its position relative to the colons, not from how many expressions there
+  // are (a[:2] is a stop, a[::2] is a step). A fully open slice like a[:] has
+  // no expressions at all.
+  let startExpression: ExpressionNode | null = null;
+  let stopExpression: ExpressionNode | null = null;
+  let stepExpression: ExpressionNode | null = null;
+  const functionDeclarations: FunctionDeclarationNode[] = [];
+
+  for (const expressionCtx of ctx.expression_list()) {
+    const colonsBefore = colons.filter(
+      (colon) => colon.symbol.tokenIndex < expressionCtx.start.tokenIndex,
+    ).length;
+
+    const expression = visitor.getExpression(expressionCtx);
+    functionDeclarations.push(...expression.functionDeclarations);
+
+    if (colonsBefore === 0) {
+      startExpression = expression.node;
+    } else if (colonsBefore === 1) {
+      stopExpression = expression.node;
+    } else {
+      stepExpression = expression.node;
+    }
+  }
+
+  return createSliceExpression(
+    startExpression,
+    stopExpression,
+    stepExpression,
+    functionDeclarations,
+  );
 };
 
 const createSliceExpression = (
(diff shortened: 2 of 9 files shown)


backend/src/ast/types/general-ast/ast-nodes/statement-node/function-declaration-node.ts
```diff
@@ -13,5 +13,8 @@ export interface FunctionDeclarationNode extends StatementNodeBase {
 
   parameterNames: string[];
   decorators?: ExpressionNode[];
+  // set (to true) only for declarations the language marks asynchronous,
+  // e.g. python's `async def`
+  isAsync?: boolean;
```
pierluca: at this stage, it's not desirable to make this distinction between `async def` and `def`.
Keep it simpler.

backend/src/ast/converters/python/nodes/statements/function-def-raw.ts
```diff
@@ -26,7 +26,8 @@ export const convertFunctionDefRaw = (
   returnType: ExpressionNode | null;
   body: StatementSequenceNode;
 } => {
-  const isAsync = ctx.ASYNC() !== undefined;
+  // ANTLR's accessor returns null - not undefined - when `async` is absent
+  const isAsync = ctx.ASYNC() ? true : false;
```
pierluca:
```suggestion
  const isAsync = !!ctx.ASYNC();
```
¨but I believe this could be removed entirely, see other comment.

backend/src/ast/converters/python/nodes/expressions/yield-expr.ts
```diff
@@ -12,11 +12,11 @@ export const convertYieldExpr = (
   visitor: IPythonAstVisitor,
   ctx: Yield_exprContext,
 ): PythonVisitorReturnValue => {
-  const expression = visitor.getExpression(
-    ctx.expression() ?? ctx.star_expressions(),
-  );
+  // a bare `yield` has no value at all; the null check must happen before
+  // getExpression, which cannot handle a null context
```
pierluca:
```suggestion
```

backend/src/ast/converters/python/nodes/expressions/slice.ts
```diff
@@ -13,48 +13,45 @@ export const convertSlice = (
...
+  const functionDeclarations: FunctionDeclarationNode[] = [];
 
-  if (expressions.length > 0) {
-    let startExpression: ExpressionNode | null = null;
-    let stopExpression: ExpressionNode | null = null;
-    let stepExpression: ExpressionNode | null = null;
+  for (const expressionCtx of ctx.expression_list()) {
+    const colonsBefore = colons.filter(
+      (colon) => colon.symbol.tokenIndex < expressionCtx.start.tokenIndex,
+    ).length;
```
pierluca, line 36: does this still work even if there are nested expressions within the slice indexing?

backend/src/ast/converters/python/nodes/expressions/slice.ts
```diff
@@ -13,48 +13,45 @@ export const convertSlice = (
...
+    if (colonsBefore === 0) {
+      startExpression = expression.node;
+    } else if (colonsBefore === 1) {
+      stopExpression = expression.node;
     } else {
-      throw new Error(
-        `Unexpected slice format. Found ${colonList.length} colons, ${expressions.length} expressions and the '${ctx.start.text}' starting token.`,
-      );
+      stepExpression = expression.node;
     }
```
pierluca: `else if (colonsBefore === 2)` would be reasonable, with the actual `else` case being reserved for an exception.

backend/src/ast/converters/python/nodes/expressions/kwarg-or-starred.ts
```diff
@@ -1,37 +1,29 @@
...
+  // null here; visiting the starred expression produces the same unpack
+  // operator as a plain `f(*args)` call.
   return {
     name: null,
-    expression: {
-      nodeType: AstNodeType.expression,
-      expressionType: ExpressionNodeType.operator,
-      operator: starArgsOperator,
-      operands: [expression.node],
-    } satisfies OperatorNode,
```
pierluca, line 35: to be checked: can the new version still correctly produce this form, when the data allows it?

pierluca (commented):
Commit c383e1cf45f0cb24a9847831cda39b0451dfc99c needs to be removed.

backend/src/ast/types/general-ast/ast-nodes/statement-node/function-declaration-node.ts
pierluca: Done in 84eb5733 — dropped the distinction entirely: no `isAsync` field on `FunctionDeclarationNode`, no marker in the converted AST, and the two tests asserting the distinction are removed. `def` and `async def` intentionally convert to the same AST.

backend/src/ast/converters/python/nodes/statements/function-def-raw.ts
pierluca: Removed entirely in 84eb5733. Verified by grep that this was its only consumer: the `isAsync` locals in for-stmt.ts / with-stmt.ts are separate and still used for their `async for` / async-with operator encoding — those are untouched.

backend/src/ast/converters/python/nodes/expressions/yield-expr.ts
pierluca: Trimmed to a single line in 84eb5733 (`// null for a bare yield, which has no value at all`) — deleting only line 16 would have left the previous line dangling mid-sentence.

backend/src/ast/converters/python/nodes/expressions/slice.ts
pierluca: Yes — verified empirically rather than by reasoning alone. Two properties make it safe: (1) `ctx.COLON_list()` returns only the slice rule's *direct* colon tokens, never colons inside nested subtrees; (2) the comparison uses the bound expression's start token index, so anything inside that expression is irrelevant. Ran the adversarial cases through the converter — all bucket correctly, including the two designed to confuse a colon counter:

- `a[b[1]:c[2]]` → `create-slice-start-stop`
- `a[x + 1 : y * 2 : z ** 2]` → `create-slice-start-stop-step`
- `a[{1: 2}[1]:]` (dict literal's own colon inside the bound) → `create-slice-start`
- `a[b[1:2]:]` (nested slice, whose colons belong to the inner SliceContext) → `create-slice-start`
- `a[f(1):g(2)]` → `create-slice-start-stop`

All five are now committed as tests in `slices.spec.ts` (84eb5733).

backend/src/ast/converters/python/nodes/expressions/slice.ts
pierluca: Done in 84eb5733 — explicit `else if (colonsBefore === 2)` with the `else` throwing `Unexpected slice format`.

backend/src/ast/converters/python/nodes/expressions/kwarg-or-starred.ts
pierluca: Checked empirically — yes, with one nuance. When the data allows it (`f(x=2, *args)`), the new version produces an operator-wrapped node exactly like this one, but with the `unpackOperator` (`"*"`) instead of `star-args`: `{operator: "*", operands: [args]}` — verified byte-identical to the node a plain `f(*args)` produces via the args path, which is the consistency the analysis wants (same construct, same AST, regardless of parse route).

The old `star-args` form itself could never actually be produced: this branch only runs for the `starred_expression` grammar alternative, where `ctx.expression()` is null — so `getExpression(null)` threw before reaching this return on every input that selected it. No existing AST can contain `star-args`, hence nothing regresses; the constant is now unused (left in operators.ts — say the word if you'd like it deleted).

bubudetp (approved):

PR #669, opened 2026-08-04
Prevent invalid syntax from flooding jupyter

bubudetp:
## Summary by cubic
Handle invalid input during solution-to-AST conversion to stop re-analysis loops that flood Jupyter and spam Sentry. We detect invalid Python, malformed JSON, wrong JSON root shapes, and unsupported (task, mime) tuples; then mark the solution hash as permanently failed and skip analysis.

- **Bug Fixes**
  - Worker returns InvalidInput with error details for invalid Python, malformed JSON, invalid JSON root structure (Jupyter/Scratch), or unsupported (task, mime); analysis logs at debug level, sets failedAnalyses to permanentlyFailedAnalysisCount, and returns null to halt retries.
  - Cron jobs now consume analysis rejections so a single failure doesn’t abort the batch, preventing repeated conversions and unhandled rejections.

- **Refactors**
  - Introduced `ConversionError`, `SolutionConversionResult`, and `SolutionConversionStatus`; Jupyter converter and `PythonSyntaxError` now throw `ConversionError`; `AstConversionService.convertSolutionToAst` returns `SolutionConversionResult`.
  - Centralized `maximumNumberOfAnalysisRetries` and `permanentlyFailedAnalysisCount`; added tests for constants, worker (including malformed JSON and JSON root validation), cron, and analysis behavior.

<sup>Written for commit 6f40d2f5f687db399c59ae51621cdc0c489b0246. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/crt25/collimator/pull/669?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/src/api/solutions/solution-analysis.service.ts
+++ b/backend/src/api/solutions/solution-analysis.service.ts
@@ -1,11 +1,11 @@
-import { Injectable } from "@nestjs/common";
+import { Injectable, Logger } from "@nestjs/common";
 import { AstVersion, Prisma, Solution, SolutionAnalysis } from "@prisma/client";
 import { AstConversionService } from "src/ast/ast-conversion.service";
 import { SolutionConversionStatus } from "src/ast/converters/solution-conversion-result";
 import { PrismaService } from "src/prisma/prisma.service";
 import { incrementFailedAnalysis } from "@prisma/client/sql";
 import { TasksService } from "../tasks/tasks.service";
-import { maximumNumberOfAnalysisRetries } from "./solution-analysis.constants";
+import { permanentlyFailedAnalysisCount } from "./solution-analysis.constants";
 
 export type SolutionAnalysisCreateInput = Omit<
   Prisma.SolutionAnalysisUncheckedCreateInput,
@@ -14,6 +14,8 @@ export type SolutionAnalysisCreateInput = Omit<
 
 @Injectable()
 export class SolutionAnalysisService {
+  private readonly logger = new Logger(SolutionAnalysisService.name);
+
   constructor(
     private readonly prisma: PrismaService,
     private readonly tasksService: TasksService,
@@ -32,10 +34,14 @@ export class SolutionAnalysisService {
         solution,
       );
 
-      if (conversion.status === SolutionConversionStatus.InvalidSyntax) {
-        // invalid syntax is an expected state while a student is editing
+      if (conversion.status === SolutionConversionStatus.InvalidInput) {
+        // invalid input is an expected state while a student is editing
+        this.logger.warn(
+          `Skipping analysis for invalid solution (task id: ${solution.taskId}, hash: ${Buffer.from(solution.hash).toString("hex")}): ${JSON.stringify(conversion.errors)}`,
+        );
+
         await this.prisma.solution.update({
-          data: { failedAnalyses: maximumNumberOfAnalysisRetries },
+          data: { failedAnalyses: permanentlyFailedAnalysisCount },
           where: {
             taskId_hash: {
               taskId: solution.taskId,
--- a/backend/src/api/solutions/solution-analysis.service.ts
+++ b/backend/src/api/solutions/solution-analysis.service.ts
@@ -1,9 +1,11 @@
 import { Injectable } from "@nestjs/common";
 import { AstVersion, Prisma, Solution, SolutionAnalysis } from "@prisma/client";
 import { AstConversionService } from "src/ast/ast-conversion.service";
+import { SolutionConversionStatus } from "src/ast/converters/solution-conversion-result";
 import { PrismaService } from "src/prisma/prisma.service";
 import { incrementFailedAnalysis } from "@prisma/client/sql";
 import { TasksService } from "../tasks/tasks.service";
+import { maximumNumberOfAnalysisRetries } from "./solution-analysis.constants";
 
 export type SolutionAnalysisCreateInput = Omit<
   Prisma.SolutionAnalysisUncheckedCreateInput,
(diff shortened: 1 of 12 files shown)


backend/src/ast/converters/solution-conversion-result.ts
```diff
@@ -0,0 +1,14 @@
...
+export enum SolutionConversionStatus {
+  Success = "success",
+  InvalidSyntax = "invalid-syntax",
+}
+
+export type SolutionConversionResult =
+  | { status: SolutionConversionStatus.Success; ast: GeneralAst }
+  | {
+      status: SolutionConversionStatus.InvalidSyntax;
+      errors: PythonSyntaxErrorDetail[];
```
pierluca: This is a generic converter. It shouldn't support _only_ Python errors, they could be Scratch's too. We should have a generic `ConversionErrorDetail[]` , of which `PythonSyntaxErrorDetail` could be a subtype

backend/src/ast/converters/solution-conversion-worker.piscina.ts
```diff
@@ -19,29 +24,40 @@ const SolutionConversionWorker = ({
...
+        taskType === TaskType.JUPYTER &&
+        solution.mimeType === "application/json"
+      ) {
+        ast = convertJupyterToGeneralAst(
+          JSON.parse(new TextDecoder("utf-8").decode(solution.data)),
+        );
+      } else {
+        throw new Error(
+          `Unsupported (task, solution mime type) tuple '(${taskType}, ${solution.mimeType})'`,
+        );
```
pierluca: This should likely be its own `return { status: InvalidThingy, ... }`

backend/src/ast/converters/solution-conversion-worker.piscina.ts
```diff
@@ -19,29 +24,40 @@ const SolutionConversionWorker = ({
...
+      }
+
+      return { status: SolutionConversionStatus.Success, ast };
+    } catch (error) {
+      if (error instanceof PythonSyntaxError) {
```
pierluca: We're in a generic worker, as per the other comment, this shouldn't _only_ support  `PythonSyntaxError`s

backend/src/ast/ast-conversion.service.ts
```diff
@@ -35,7 +35,7 @@ export class AstConversionService implements OnModuleDestroy {
   async convertSolutionToAst(
     task: TaskWithoutData,
     solution: Solution,
-  ): Promise<GeneralAst> {
+  ): Promise<SolutionConversionResult> {
```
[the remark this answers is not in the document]
pierluca: Updating the JSDoc wouldn't hurt, indeed :)

backend/src/api/solutions/solution-analysis.service.ts
```diff
@@ -21,16 +23,32 @@ export class SolutionAnalysisService {
...
+          data: { failedAnalyses: maximumNumberOfAnalysisRetries },
+          where: {
+            taskId_hash: {
+              taskId: solution.taskId,
+              hash: solution.hash,
+            },
+          },
+        });
+
+        // resolve normally so sentry does not receive an unhandled rejection
```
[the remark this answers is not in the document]
pierluca: This is a good point, we should probably log them here.

pierluca (changes requested):

backend/src/api/solutions/solution-analysis.service.ts
```diff
@@ -21,16 +23,32 @@ export class SolutionAnalysisService {
...
+      const conversion = await this.astConversionService.convertSolutionToAst(
         task,
         solution,
       );
 
-      const genericAst = JSON.stringify(ast);
+      if (conversion.status === SolutionConversionStatus.InvalidSyntax) {
+        // invalid syntax is an expected state while a student is editing
+        await this.prisma.solution.update({
+          data: { failedAnalyses: maximumNumberOfAnalysisRetries },
```
pierluca: **Claude:** Design nit: failedAnalyses = 3 overloads a retry counter as a permanent "invalid" flag. It works because solution hashes are immutable, but if the constant is ever raised, previously-flagged solutions silently resurrect into the retry loop.

**Pierluca:** an ugly but effective solution would be to have a second constant along the lines of `permanentlyFailedRetryCount = 9999`, marked as "never change this" and have a unit test that checks (for future sanity) that `maximumNumberOfAnalysisRetries < permanentlyFailedRetryCount`

backend/src/api/solutions/solution-analysis.service.ts
```diff
@@ -21,16 +23,32 @@ export class SolutionAnalysisService {
...
 
     try {
-      const ast = await this.astConversionService.convertSolutionToAst(
+      const conversion = await this.astConversionService.convertSolutionToAst(
         task,
         solution,
       );
 
-      const genericAst = JSON.stringify(ast);
+      if (conversion.status === SolutionConversionStatus.InvalidSyntax) {
```
pierluca: **Claude:** No test covers this new path. A cheap unit test — worker returns InvalidSyntax → service updates failedAnalyses and resolves null — would lock this in.

backend/src/api/solutions/solution-analysis.service.ts
pierluca, line 1: **Claude:** Worth folding into this PR since its goal is stopping Sentry spam: at solutions.service.ts:704 the cron does .performAnalysis(...).catch() with no handler, under a comment saying "ignore exceptions, we'll just re-try". A no-arg .catch() does not swallow the rejection — it passes it through, so Promise.all rejects and the cron method throws to Sentry on every failed batch. It should be .catch(() => {}) (both cron sites). This is plausibly a chunk of the very spam this PR is chasing.

pierluca (changes requested):
Added feedback from Claude, which spotted a few extra things absolutely worth fixing.

pierluca (approved):

PR #670, opened 2026-08-04
[CRT-379] Handle replayed student activities

bubudetp:
## Summary by cubic
Make student activity creation idempotent to stop crashes from replayed client events (CRT-379). Simplified retry flow: on unique conflicts, return the existing activity or retry once for concurrent writes.

- **Bug Fixes**
  - Treat (`studentId`, `type`, `happenedAt`, `happenedAtCounter`) as an idempotency key; on `P2002`, find via `uniqueStudentActivityPerTypeAndTime`, return existing, and skip analysis.
  - Add `happenedAtCounter` to `StudentActivityInput` and the unique lookup.
  - Replace transactional batch creation with per-item `create()` so idempotency and a single retry run per activity; rethrow persistent conflicts; add tests for replays and concurrent solution races.

<sup>Written for commit 1cb98bc57ec4c909e8840d0b2bb0f1fdffe8d0df. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/crt25/collimator/pull/670?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/src/api/student-activity/student-activity.service.ts
+++ b/backend/src/api/student-activity/student-activity.service.ts
@@ -20,8 +20,9 @@ export type AppActivityInput = Omit<
 
 export type StudentActivityInput = Omit<
   Prisma.StudentActivityUncheckedCreateInput,
-  "solutionHash" | "appActivity" | "studentId"
+  "solutionHash" | "appActivity" | "studentId" | "happenedAtCounter"
 > & {
+  happenedAtCounter: number;
   appActivity: AppActivityInput | null;
 };
 
@@ -40,20 +41,11 @@ export class StudentActivityService {
       solution: SolutionInput;
     }[],
   ): Promise<StudentActivity[]> {
-    const results = await this.prisma.$transaction(
-      activityWithSolution.map(({ activity, solution }) =>
-        this.prisma.studentActivity.create({
-          data: this.buildActivityInput(student, activity, solution),
-          include: { solution: true },
-        }),
-      ),
-    );
+    const results: StudentActivity[] = [];
 
-    results.forEach((result) =>
-      // do not wait for the promise to resolve
-      // this will happen in the background
-      this.analysisService.performAnalysis(result.solution, latestAstVersion),
-    );
+    for (const { activity, solution } of activityWithSolution) {
+      results.push(await this.create(student, activity, solution));
+    }
 
     return results;
   }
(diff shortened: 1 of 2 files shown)


backend/src/api/student-activity/student-activity.service.ts
pierluca, line 158: **Claude:** `buildActivityInput` uses `connectOrCreate` for the solution, which has the well-known concurrent-race P2002 (two simultaneous requests carrying the same new solution but different activities). In that case the activity-key lookup misses and the original error is rethrown, failing a request that a single retry of the create would have satisfied. Consider inspecting error.meta.target to distinguish which constraint fired, or retrying the create once when the activity lookup comes up empty. Pre-existing race, but this PR is the natural home for it.

backend/src/api/student-activity/student-activity.service.ts
```diff
@@ -63,10 +55,46 @@ export class StudentActivityService {
...
+          },
+        },
+        include: { solution: true },
+      });
+
+      if (!existingActivity) {
+        throw error;
+      }
+
+      return existingActivity;
```
pierluca: Technically this could be returning a different solution than the one we submitted, since we're not comparing the submitted solution with the existing one.

I don't think it's a problem, but perhaps add a comment to mark it as an accepted trade-off between complexity and correctness?

pierluca: **Claude:** No test for the replay path; a service spec asserting "P2002 → returns existing row, no analysis triggered" would be cheap.

pierluca (changes requested):
Mostly LGTM, tiny remarks below.

backend/src/api/student-activity/student-activity.service.ts
```diff
@@ -40,20 +41,11 @@ export class StudentActivityService {
...
-    );
-
-    results.forEach((result) =>
-      // do not wait for the promise to resolve
-      // this will happen in the background
-      this.analysisService.performAnalysis(result.solution, latestAstVersion),
-    );
+    const results: StudentActivity[] = [];
+
+    for (const { activity, solution } of activityWithSolution) {
```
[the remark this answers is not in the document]
pierluca: Yes this can lead to N sequential round-trips, but partial writes is what makes the system work.  Acceptable.

pierluca (approved):

PR #677, opened 2026-08-05
[CRT-463] Forward all Nest logs to Sentry

pierluca:
## Summary

Sends every NestJS log call — all levels — to Sentry, changing only the central logging configuration (Jira: CRT-463).

- `SentryLogger` (new, `backend/src/sentry.logger.ts`) extends Nest's `ConsoleLogger`: each level method forwards to the official Sentry **Logs** API (`Sentry.logger.trace/debug/info/warn/error/fatal`) and then delegates to `super`, so console formatting and the `LOG_LEVEL` filtering behave exactly as before.
- `instrument.ts` gains `enableLogs: true` (supported by our `@sentry/nestjs` 10.32.1).
- `main.ts` passes a `SentryLogger` instance to `NestFactory.create` with the same level filter as before — the only call-site change in the codebase.

## Why not the article's / console-integration approach

- The referenced dev.to article (custom logger + `captureMessage` per log) predates the official Sentry Logs product; capturing every log as a Sentry *event* would flood the issue stream. The subclass-the-logger shape is kept; the transport is the modern `Sentry.logger.*` API.
- Sentry's `consoleLoggingIntegration` patches `console.*`, but Nest's `ConsoleLogger` writes via `process.stdout.write`, so it would silently miss all application logs.

## Behavior notes

- Sentry deliberately receives **all** levels, including those the console suppresses (`LOG_LEVEL` unset → console still prints only error/warn/fatal).
- With no `SENTRY_DSN` (development, tests) the forwarding calls are no-ops.
- Error calls keep their stack: `error(message, stack, context)` and logged `Error` objects both attach `stack` and `context` as log attributes.

## Verification

- New unit spec `src/__tests__/sentry.logger.spec.ts` (11 tests, written red-first): level mapping, console delegation, forwarding of console-suppressed levels, stack/context extraction, non-string messages.
- Full backend unit suite green (the two `PrismaClientInitializationError` failures locally are the known DB-backed specs needing a reachable Postgres; they pass against one).
- Boot smoke test: app builds, starts with the new logger, console output byte-identical, endpoints serving.

---
## Summary by cubic
Forward all NestJS logs (every level) to Sentry Logs via a new `SentryLogger`, keeping console output and `LOG_LEVEL` behavior unchanged. Sentry log attributes now exactly match Nest’s console by reusing `ConsoleLogger`’s context/stack classification (CRT-463).

- **New Features**
  - `SentryLogger` extends `ConsoleLogger` and mirrors its public overloads; forwards to `Sentry.logger.trace/debug/info/warn/error/fatal`, then logs to console unchanged.
  - App boot uses `SentryLogger` with the same level filter; all levels still go to Sentry even if the console suppresses them; with no `SENTRY_DSN`, forwarding is a no-op.
  - Enabled Sentry Logs with `enableLogs: true` in `@sentry/nestjs` init.
  - Context/stack parsing uses copies of `ConsoleLogger`’s helpers for exact parity: only `error()` may attach a stack; a single string is a stack only if it looks like one; trailing string is the context; falls back to the logger’s configured context.

<sup>Written for commit c983ad92e4a1a20a54d66ae84145c6fe4b845f66. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/crt25/collimator/pull/677?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/src/sentry.logger.ts
pierluca, line 1: This should be reviewed somewhat side-by-side with:
https://github.com/nestjs/nest/blob/master/packages/common/services/console-logger.service.ts

bubudetp (approved):

PR #679, opened 2026-08-06
[CRT-454] Report a task-started activity when a student opens a task

pierluca:
## Summary

An anonymous student was invisible in the teacher's progress view until they submitted something, because that view derives its participants from analysed solutions (Jira: CRT-454). Fixed per review by having the **app report opening the task** — no session-API surface, and the solution comes from the app in its own encoding.

## Approach (app-push)

- New `postTaskStarted` app→platform RPC carrying the solution the task was opened with. The platform records it as a **TASK_STARTED** activity (`Task.tsx`, beside the existing `onSolutionRun` → TASK_RUN_SOLUTION). The analysis cron then analyses that solution and the student surfaces in the progress view — the exact path a solution-run already takes. It also gives replay a starting point.
- **The frontend never sources the solution.** As you noted, the platform's stored task/submission files aren't interchangeable with a solution (a Jupyter submission is only the student notebook, not the whole task). So each app emits its *own* raw encoding:
  - **Scratch** → `vm.toJSON()`, exactly what `postSolutionRun` sends (no assertion run).
  - **Jupyter** → the raw student notebook file (not the otter-graded `getSubmission`, which would hang the kernel).
- Emitted once the task has loaded for a **solving** student (`isStudentSolving` / `Mode.solve`) and again when the task is re-presented in another language; edit and show modes never emit.

## Why not the session API (previous version)

Reverted entirely — exposing participants on the session DTO treated anonymous students as a special case and had the frontend handling solution data. This version needs no session/progress change: joining plus opening a task makes the student visible through the existing analysed-activity path.

## Latent bug fixed along the way

Sending an app→platform message *during* a load handler revealed that the app's RPC origin was only applied by a React state effect that runs *after* the handler returns — so the first message was silently dropped. `useIframeParent` now sets the origin synchronously too (Jupyter's own init already did).

## Locale note

Re-emitting on language change is included (a fresh TASK_STARTED per locale). Persisting *which* locale would need a `StudentActivity` column; deferred until replay needs it — say the word.

## Verification

- New Scratch e2e proving the app emits `postTaskStarted` (with a solution) on load — red→green. Full apps/scratch suite green (101 passed; the shared `loadTask` helper and three positional assertions were made robust — find-by-method — since a solving load now legitimately posts one extra message).
- Frontend jest 142/142, notebook-runner 57/57, tsc/eslint/prettier clean across libraries + both apps + frontend.
- Backend unchanged: `TASK_STARTED` is already in the enum and `TrackStudentActivityDto`, and `solutions.queries.spec.ts` already proves an activity-sourced solution surfaces the student.


libraries/iframe-rpc-react/src/useIframeParent.ts
```diff
@@ -47,6 +47,10 @@ export const useIframeParent = (
               // cache the parent origin and set it in state
               cachedParentOrigin = args[1].origin;
               setTaskOrigin(args[1].origin);
+              // also set it synchronously, so a request the handler sends back
+              // to the platform (e.g. postTaskStarted) has a live origin - the
+              // state effect only runs after the handler has returned
+              crtPlatform.current.setOrigin(args[1].origin);
```
pierluca: the comment is not great, but this is the actual bugfix.
Without this, the origin is set too late, and the response doesn't go through.

@bubudetp what comment would you suggest having here?

bubudetp: this is important to document well, "live origin" is too vague to me, it can also benefit from a rewritten version but in plain english & with more context

here's my suggestion: 
```suggestion
              // the load handler may start a new request from the app to the platform such as postTaskStarted
              // react only applies the taskOrigin in a later effect, so we set the origin synchronously before invoking the handler
              crtPlatform.current.setOrigin(args[1].origin);
```

bubudetp (approved):

PR #688, opened 2026-08-10
[CRT-339] Show submitted solutions only in the analysis dashboard

pierluca:
### Root cause

Test pass/fail results are stored only on graded submissions (`SolutionTest.studentSolutionId`). The analysis dashboard, however, selects **per student the single most-recent record across both `StudentSolution` and `StudentActivity`** (`getCurrentAnalysesWithActivities.sql`) and joins the tests off *that exact row's* `studentSolutionId`.

Scratch continuously emits testless `StudentActivity` rows — every green-flag run (`postSolutionRun`) and every block edit. As soon as one of those is newer than the graded submission, the dashboard lands on a row with `studentSolutionId = NULL`, the `SolutionTest` join yields nothing, and the **"Passed Tests"** axis reads `0`. The submission's tests are stored correctly (the submit toast even shows the right count) — they are just shadowed on read-back.

Jupyter emits no activities (it only auto-saves on a dirty edit), so its newest record is normally the graded submission — which is why it was unaffected.

### Fix

The analysis dashboard is only consulted once students are done, so it should show **submitted solutions only** and never union in student activities (per product decision).

- New submitted-only query variants (`getCurrentStudentSolutionAnalyses` / `getSoftDeletedCurrentStudentSolutionAnalyses`) that are identical to the existing `…WithActivities` queries but **do not union `StudentActivity`** — so a student's latest *submitted* solution, and its test results, is always the row shown. Test-less submitted solutions are still shown.
- Exposed via a new **optional** `studentSolutionsOnly` query param on `GET …/current-analyses` (defaults to off).
- Only the analysis dashboard (`Analyzer`) opts in. **Every other consumer of the endpoint is unchanged** (Showcase, Dissimilar\*, ProgressList, per-student page, task-instance progress).

### Scope / blast radius

- `studentSolutionsOnly` defaults to `false`, so the shared `useCurrentSessionTaskSolutions` hook and all its other callers issue byte-identical requests.
- The three reference-solution branches (task references, starred submissions) are preserved; only the `StudentActivity` sources were removed from the submitted-only variants.

### Verification status

- ✅ Frontend `tsc --noEmit` clean for the changed files; Prettier-formatted.
- ⚠️ **Backend was not built/tested in this environment** (the dev Postgres runs inside WSL/Docker and is not reachable from the host session; Prisma `typedSql` needs a live DB). Before merge, in a DB-enabled environment:
  1. `cd backend && yarn prisma:generate` — regenerates typed-SQL for the two new `.sql` files.
  2. `yarn build && yarn api:generate` then `cd ../frontend && yarn update:api` — regenerates the OpenAPI spec + orval client. (`backend/docs/api.json` and the generated params model were updated by hand to match; this step should reproduce them 1:1.)
  3. `cd backend && yarn test` — runs the added `getCurrentStudentSolutionAnalyses` query integration tests.

### Tests added

`solutions.queries.spec.ts` covers the submitted-only query: latest submission wins over a newer activity (with tests preserved), test-less submissions still returned, activity-only students excluded, and task reference solutions still returned.

Closes CRT-339.

---
## Summary by cubic
Fixes CRT-339 by making the analysis dashboard show only submitted solutions and ignore past starred rows so Scratch passed-test counts are correct. Star/unstar now refreshes all `current-analyses` cache variants via a shared key matcher so the dashboard reflects changes immediately.

- **Bug Fixes**
  - SQL: Added $3 `studentSolutionsOnly` and $4 `ignoreStarredSolutions` to `getCurrentAnalysesWithActivities.sql` (+ soft-deleted) to drop activities and past starred rows; the latest starred solution still returns. Clarified param docs.
  - API: Exposed optional `studentSolutionsOnly` and `ignoreStarredSolutions` on GET `/current-analyses`. Service unified as `findCurrentAnalyses(sessionId, taskId, { includeSoftDeleted?, studentSolutionsOnly?, ignoreStarredSolutions? })`.
  - Frontend: `Analyzer` requests `{ studentSolutionsOnly: true, ignoreStarredSolutions: true }`. `useCurrentSessionTaskSolutions` now accepts `{ config?, params? }` and forwards params; `TaskInstanceProgressList` updated. Introduced shared `matchesCurrentAnalysesKey` and used it in `usePatchStudentReferenceSolutions` and `useRevalidateSolutionList` to optimistically mutate and revalidate all `current-analyses` SWR keys by prefix (using the updater form).
  - Tests: Added coverage for submitted-only behavior, ignoring past starred solutions, keeping latest starred with tests, excluding activity-only students, param forwarding, the shared key matcher, and cache-key prefix mutation; updated service specs for the new options shape; added hook tests.

<sup>Written for commit 1475b3c870b4c602dce4a432038fed6fe6cfda3c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/crt25/collimator/pull/688?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


frontend/src/api/collimator/hooks/solutions/useCurrentSessionTaskSolutions.ts
```diff
@@ -17,13 +18,14 @@ export const fetchSolutionsAndTransform = (
   classId: number,
   sessionId: number,
   taskId?: number,
+  params: SolutionsControllerFindCurrentAnalysesV0Params = {},
```
pierluca, line 26: should we define our own structure here, to decouple frontend from backend? 🤔 
In theory, yes, in practice I'm not sure it's worth the effort given it's only used in one place at the moment and we can only do it when we need it (YAGNI).

bubudetp (approved):

